PR #14, opened 2022-10-13
Add `useLanguage` hook and `withTranslation` hoc

### LINKED ISSUE #9: Internationalization hooks

MaxSchaefer:
#### `useLanguage`
This hook provides the storage and methods to pass a language context to the app.

#### `withTranslation`
This hoc enables components to register defaultTranslations (_wording wip_) while the actual components remains independent of the text and translations.

#### Examples

There are some examples with the new componets in `/pages/intl`.

Closes #9


app/pages/intl/Title.tsx
```diff
@@ -0,0 +1,38 @@
...
+    page: (page) => `Page ${page}`
+  },
+  [Language.DE]: {
+    welcome: 'Willkommen',
+    goodToSeeYou: 'Schön dich zu sehen',
+    page: (page) => `Seite ${page}`
+  }
+}
+
+export default withTranslation(Title, defaultTitleTranslations)
```
JannikWibker, line 38: This file should not be in the `pages` folder, but rather in a `components` folder!
`/intl/Title` shouldn't be a navigable route.

app/pages/intl/test.tsx
```diff
@@ -0,0 +1,17 @@
...
+    <ProvideLanguage>
+      <div className={'flex justify-center items-center h-screen'}>
+        <Title name={'Testine Test'} />
+        <LanguageSwitcher/>
+      </div>
+    </ProvideLanguage>
+  )
+}
+
+export default Test
```
JannikWibker, line 17: Should most likely move this to `app/pages/intl.tsx`
As this is only a demo for now it isn't all that important though.

app/pages/intl/LangugeSwitcher.tsx
```diff
@@ -0,0 +1,33 @@
...
+const defaultLanguageSwitcherTranslations: Translations<LanguageSwitcherTranslation> = {
+  [Language.EN]: {
+    toggleLanguage: 'Toggle language',
+  },
+  [Language.DE]: {
+    toggleLanguage: 'Sprache wechseln',
+  }
+}
+
+export default withTranslation(LanguageSwitcher, defaultLanguageSwitcherTranslations)
```
JannikWibker, line 33: This file should not be in the `pages` folder, but rather in a `components` folder!
`/intl/LanguageSwitcher` shouldn't be a navigable route.

app/hooks/useLanguage.tsx
```diff
@@ -0,0 +1,38 @@
...
+  EN = 'en-US',
+  DE = 'de-DE'
+}
+
+export const DEFAULT_LANGUAGE = Language.DE
+
+export interface LanguageContextValue {
+  language: Language,
+  setLanguage: Dispatch<SetStateAction<Language>>
+}
```
JannikWibker, line 14: Should probably decide on using either `type` or `interface` for most things, I personally much prefer `type` as it gives a lot more freedom with things like union and intersection types.

app/hocs/withTranslation.tsx
```diff
@@ -0,0 +1,17 @@
+import type { ComponentType } from 'react'
+import type { Language } from '../hooks/useLanguage'
+import { useLanguage } from '../hooks/useLanguage'
+
+export type EventualString = string | Promise<string> | ((...args: any[]) => (string | Promise<string>));
```
JannikWibker, line 5: I'm unsure about allowing promises here, are there any concrete use cases where that would be nice to have?

app/hocs/withTranslation.tsx
```diff
@@ -0,0 +1,17 @@
...
+export type Translation<T = Record<string, EventualString>> = Record<keyof T, EventualString>;
+export type Translations<T = Translation> = Record<Language, T>;
+export type PropsWithTranslation<T = Translation, P = unknown> = P & { translation: T };
```
JannikWibker, line 8: I'm uncertain what benefit a default value/type brings us here instead of constraining the input type.
I'd change this to `<T extends [X]>` where `[X]` is either `Record<string, EventualString>` or `Translation`.

app/pages/intl/test.tsx
```diff
@@ -0,0 +1,17 @@
+import type { NextPage } from 'next'
+import Title from './Title'
+import LanguageSwitcher from './LangugeSwitcher'
+import { ProvideLanguage } from '../../hooks/useLanguage'
+
+const Test: NextPage = () => {
+  return (
+    <ProvideLanguage>
+      <div className={'flex justify-center items-center h-screen'}>
```
JannikWibker, line 9: You can use just straight old `className="flex justify-center items-center h-screen"`; no need for `{' ... '}`

app/pages/intl/LangugeSwitcher.tsx
```diff
@@ -0,0 +1,33 @@
+import type { FunctionComponent } from 'react'
+import type { PropsWithTranslation, Translation, Translations } from '../../hocs/withTranslation'
+import { withTranslation } from '../../hocs/withTranslation'
+import { Language, useLanguage } from '../../hooks/useLanguage'
+
+interface LanguageSwitcherTranslation extends Translation {
+  toggleLanguage: string
+}
```
JannikWibker, line 8: `type` vs `interface` again.
Is there any practical use for the `extends Translation` here?
With the change suggested in `app/hocs/withTranslation.tsx` this would just be a type error if it didn't conform to the Translation type when used further below with `Translations<...>`.

app/pages/intl/test.tsx
MaxSchaefer: Should we enforce this with the following eslint rules?
`'react/jsx-curly-brace-presence': ['error', 'never']`
`'jsx-quotes': ['error', 'prefer-single']`

app/hocs/withTranslation.tsx
MaxSchaefer: I thought about it, no there doesn't seem to be any real use cases since the translations are mostly used directly in TSX.

app/pages/intl/Title.tsx
MaxSchaefer: Its just an example. _Moved!_

app/pages/intl/test.tsx
MaxSchaefer: _Moved!_

app/pages/intl/LangugeSwitcher.tsx
MaxSchaefer: _Moved!_

app/pages/intl/test.tsx
JannikWibker: Yes this sounds like a good idea

There are probably still some more rules which are useful but missing, in such a case I'd suggest just quickly asking around if anyone's opposed to the rule and if not adding it real quick.

app/components/examples/Title.tsx
```diff
@@ -0,0 +1,38 @@
...
+  name: string
+}
+
+// Simple Title component to demonstrate some translations
+const Title: FunctionComponent<PropsWithTranslation<TitleTranslation, TitleProps>> = (props) => {
+  return <p
+    className={'rounded bg-gray-800 text-gray-200 p-1 px-2'}
+  >{props.translation.welcome}! {props.translation.goodToSeeYou}, <span
+    className={'text-green-300'}
+  >{props.name}</span>. {props.translation.page(123)}</p>
```
JannikWibker, line 22:
```suggestion
  return (
    <p className="rounded bg-gray-800 text-gray-200 p-1 px-2">
      {props.translation.welcome}{'! '}
      {props.translation.goodToSeeYou}{', '}
      <span className="text-green-300">{props.name}</span>{'. '}
      {props.translation.page(123)}
    </p>
  )
```
*There's probably an eslint rule for this*

app/.eslintrc.js
```diff
@@ -52,8 +53,19 @@ module.exports = {
         }
       }
     ],
+    'jsx-quotes': ['error', 'prefer-double'],
```
MaxSchaefer, line 56: @JannikWibker should we really define both types of quotes (take a look at line #18)?

app/hocs/withTranslation.tsx
```diff
@@ -0,0 +1,21 @@
+import type { ComponentType } from 'react'
+import type { Language } from '../hooks/useLanguage'
+import { useLanguage } from '../hooks/useLanguage'
+
+export type EventualString = string | ((...args: any[]) => string);
+export type Translation<T = Record<string, EventualString>> = Record<keyof T, EventualString>;
+export type Translations<T extends Translation> = Record<Language, T>;
+export type PropsWithTranslation<P, T extends Translation> = P & { translation: T };
```
MaxSchaefer, line 8: I think we should extend P like react itself with the `PropsWithChildren` type
`type PropsWithChildren<P = unknown> = P & { children?: ReactNode | undefined };`
There is not always a need for props. And an `Empty` type in that case seams unnecessary.

app/hocs/withTranslation.tsx
MaxSchaefer: _Reopened  at https://github.com/helpwave/web/pull/14#discussion_r996277578_

app/.eslintrc.js
JannikWibker: The one above is for JS/TS quotes, the one below for JSX. As JSX mirrors HTML in syntax the standard is to use double quotes. Thus I'd say defining both is a good thing.

This is also the default in many other projects, even when it is not necessarily enforce by linting rules. As a side note the default in prettier is also to use double quotes in JSX.

app/hocs/withTranslation.tsx
JannikWibker: We could do `PropsWithTranslation<P = unknown, T extends Translation = Translation>` or change the order back around (I initially swapped it as I thought it would be more natural to have props come first)

app/.eslintrc.js
MaxSchaefer: Makes sense, lets stick with the standards 👍🏻

app/hocs/withTranslation.tsx
MaxSchaefer: _Refactored!_

JannikWibker (approved):

PR #34, opened 2022-12-14
feat: basic team member section

JannikWibker:
Screenshot 2022-12-14 at 20 02 26

Very very basic first implementation of the team member page.
There are a ton of things that need to be refactored or done properly later on.

Things not implemented/needed improvements:
- dotted gradients
- general refactoring
- move team members some place else maybe?
- move things to their own component files
- add some content on the left side (this needs to be designed first)
- migrate to twind? this is a general pain point of mine
- colors are all over the place (firstly from a design perspective -> should use more consistent ones maybe or standardise the ones currently in use *and* secondly from a code structure perspective -> use the tailwind config more)

### DIFF
--- a/landing-page/components/TeamSection.tsx
+++ b/landing-page/components/TeamSection.tsx
@@ -6,12 +6,12 @@ type Role = typeof roles[number]
 
 const teamMembers = [
   { name: 'Felix', role: 'Frontend Developer' },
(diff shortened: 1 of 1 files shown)


MaxSchaefer:
Nice, thanks! That was fast. I keep "Things not implemented/needed improvements"-List in mind.

MaxSchaefer (approved):

PR #41, opened 2023-01-05
start using twind instead of tailwind

### LINKED ISSUE #40: switch to twind
> reference: https://twind.style & https://github.com/tw-in-js/twind

Twind is a replacement for Tailwind which is compatible with Tailwind's classname system but generates the styles on the fly. This removes one major annoyance with Tailwind, this being not being able properly work with conditional styles.
With SSR this isn't anything all too different from just using tailwind as the backend generates styles and hands these over to the client. The important difference now is that the client can continue to generate styles for new branches taken during execution.

Getting twind working just how you want it is somewhat complicated and requires a little bit of digging through source code

JannikWibker:
Closes #40


MaxSchaefer:
Can you provide some examples in the changes of the pull request where twind is better than tailwind?

JannikWibker:
Not something currently present in the PR but there are a lot of instances where tailwind doesn't compile classnames and you are confused as to why at first.

Let's take the landing page as an example:

This hacky solution for setting the text color was needed because otherwise tailwind wouldn't compile all of the `text-[...]` classnames
image

With twind this could look like this:
```tsx
<div className={tw(`font-medium text-xl text-[${getColorByRole(role)}]`)}>{role}</div>
```

JannikWibker:
For now I've just updated the dashboard but I plan on switching the landing page over to twind as well

MaxSchaefer (approved):
Dashboard looks good.

JannikWibker:
I'll do the landing page in a separate PR

PR #47, opened 2023-01-16
SEO Optimizations

### LINKED ISSUE #16: SEO metadata, favicon
Basically make open graph link validators happy, set appropriate `<title />`-tags per page, add a favicon, manifest.json, ...

use-to:
closes #16 

- [X] `manifest.json`
- [X] `robots.txt`
- [X] proper links
- [X] image descriptions
- [X] favicon
- [X] canonical tag
- [X] general meta tags (open graph)
- [X] twitter tag

possible todos:
- add related applications in `manifest.json` (Google Play, AppStore, ...)
- more link to enable better page ranking scores
- sitemap
- more relevant text snippets
- using html headings and proper structure
- provide screenshots, previews and icons for `manifest.json`
- improve Lighthouse's `performance` and `accessibility` scores


landing-page/components/sections/PartnerSection.tsx
```diff
@@ -6,19 +6,21 @@ import Ukm from '../../icons/partners/ukm'
...
   return (
     <TitleSection id="partner" ref={ref} title="Our Partners">
       <div className={tw('flex gap-4 pt-4')}>
       {partners.map((partner) => (
-        <div key={partner.name} className={tw('py-2 px-8 bg-white rounded-lg')}>
-          <partner.Icon className={tw('w-44 h-16')} />
-        </div>
+        <a key={partner.name} href={partner.url} target="_blank" rel="noopener noreferrer">
+          <div className={tw('py-2 px-8 bg-white rounded-lg')}>
+            <partner.icon className={tw('w-44 h-16')} />
```
JannikWibker, line 21: interesting that this works, at least for components not accessed using `.` you have to use uppercase letters as lowercase ones are reserved for actual DOM elements (`div`, `img`, ..), I thought this would apply here as well

landing-page/public/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://json.schemastore.org/web-manifest-combined.json",
+  "name": "helpwave",
+  "short_name": "helpave",
```
JannikWibker, line 4: missing `w`

landing-page/public/manifest.json
```diff
@@ -0,0 +1,20 @@
...
+  "lang": "en",
+  "description": "Next generation healthcare software approach!",
+  "orientation": "portrait",
+  "categories": ["software", "medical", "health", "healthcare", "company"],
+  "scope": "https://helpwave.de/",
+  "start_url": "https://helpwave.de",
+  "background_color": "#281c20",
+  "theme_color": "#281c20",
```
JannikWibker, line 12: uncertain about these values, why was the background color chosen as the theme color as well?

use-to: The `theme_color` is the one used for the browser frame and application switcher on mobile.
The `background_color` is an experimental attribute which defines the background color before the stylesheet has loaded. [See here](https://developer.mozilla.org/en-US/docs/Web/Manifest/background_color) 
So both should use the same color (in this case the dark-red-brown tone `#281c20`).

landing-page/components/sections/PartnerSection.tsx
use-to: Okay - interesting. The linter is'nt marking that one.

landing-page/public/manifest.json
JannikWibker: works for me :+1:

landing-page/components/sections/PartnerSection.tsx
JannikWibker: if it builds and renders correctly it seems to work, I just always used an uppercase letter because I thought it doesn't work otherwise
thinking about it, I do like it more with an uppercase letter regardless of it being required or not as it states the intent of that property better

JannikWibker (approved):

JannikWibker:
We might want to add a proper (opengraph) thumbnail image as well

PR #55, opened 2023-01-23
add ger and en texts

aegis301:
Added some german texts to the landing page along with english translations. Content is open for discussion!


use-to (approved):
On the long run we should implement a language switcher #56

MaxSchaefer (approved):
Nice ✏️

PR #60, opened 2023-01-28
Intl - Internationalization

JannikWibker:
This aims to add intl / i18n / internationalization support to the landing page

What is still missing:
- [ ] a few translations here and there
- [ ] layout fixes where different translations have given elements different sizes per language
- [ ] language switching using some kind of UI
- [ ] persisting language choices

### DIFF
--- a/app/hooks/useLanguage.tsx
+++ b/app/hooks/useLanguage.tsx
@@ -1,16 +1,14 @@
-import type { Dispatch, FunctionComponent, PropsWithChildren, SetStateAction } from 'react'
 import { createContext, useContext, useEffect, useState } from 'react'
+import type { Dispatch, FunctionComponent, PropsWithChildren, SetStateAction } from 'react'
 
-export enum Language {
-  EN = 'en-US',
-  DE = 'de-DE'
-}
+const languages = ['en', 'de'] as const
+export type Languages = typeof languages[number]
 
-export const DEFAULT_LANGUAGE = Language.DE
+export const DEFAULT_LANGUAGE = 'en'
 
 export type LanguageContextValue = {
-  language: Language,
-  setLanguage: Dispatch<SetStateAction<Language>>
+  language: Languages,
+  setLanguage: Dispatch<SetStateAction<Languages>>
 }
 
 export const LanguageContext = createContext<LanguageContextValue>({ language: DEFAULT_LANGUAGE, setLanguage: (v) => v })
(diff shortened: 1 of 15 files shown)


app/hooks/useLanguage.tsx
```diff
@@ -1,39 +1,33 @@
-import type { Dispatch, FunctionComponent, PropsWithChildren, SetStateAction } from 'react'
 import { createContext, useContext, useEffect, useState } from 'react'
+import type { Dispatch, FunctionComponent, PropsWithChildren, SetStateAction } from 'react'
 
-export enum Language {
-  EN = 'en-US',
-  DE = 'de-DE'
-}
+const languages = ['en', 'de'] as const
```
MaxSchaefer, line 4: Why do we get rid of the enum?

MaxSchaefer (commented):
@JannikWibker do you think it makes sense to move the intl stuff into a common directory/lib?

JannikWibker:
Eventually yes
This has been on my mind for quite some time because there are a lot of shared things in between both `app/` and `landing-page` (twind config, some components (with slight modifications though)), but I haven't gotten around to addressing this yet.

For this we would have to think about how we do the bundling / deploying a bit: We could just import it using `workspace:../<something>` (or `file:../<something>`) or have a proper npm/github package registry release setup additionally which would allow us to import this "library" in other places (repos) as well.

app/hooks/useLanguage.tsx
JannikWibker: basically boils down to enums being an anti-pattern in typescript (somewhat)

landing-page/components/sections/ContactSection.tsx
```diff
@@ -8,8 +8,49 @@ import { Toast } from '../Toast'
...
+    yourInformation: 'DE: Your Information', // TODO: translate (in a nice sounding way; not "Ihre/Deine Informationen")
+    email: 'Email',
+    firstName: 'Vorname',
+    lastName: 'Nachname',
+    message: 'Nachricht',
+    send: 'Abschicken',
+    successMessage: 'Nachricht gesendet!',
+    failureMessage: 'Es ist ein Fehler aufgetreten!'
+  }
+}
```
use-to, line 50: Wouldn`t it make sense that we collect these translations in `.json` file for each language, so that someone can add a translations without long searching?

use-to (changes requested):

landing-page/components/sections/ContactSection.tsx
use-to: This effects more than just this few loc.

JannikWibker: I kinda want the translations to be scoped / bound to the concrete component.

Additionally there isn't only json data in the translations but also full on react components and functions and dealing with json is quite annoying when it comes to bundling and type checking

landing-page/components/sections/RoadmapSection.tsx
```diff
@@ -1,10 +1,27 @@
...
+export type RoadmapSectionLanguage = {
+  heading: string
+}
+
+const defaultRoadmapSectionTranslations: Record<Languages, RoadmapSectionLanguage> = {
+  en: {
+    heading: 'Roadmap',
+  },
+  de: {
+    heading: 'DE: Roadmap', // TODO: translate (or do we want to leave it in english as is?)
```
use-to, line 17:
```suggestion
    heading: 'Roadmap',
```
`Roadmap` is fine!

use-to (approved):

PR #85, opened 2023-02-26
chore(app): add gRPC-web example

MaxSchaefer:


app/pages/index.tsx
```diff
@@ -5,11 +5,30 @@ import { useAuth } from '../hooks/useAuth'
...
+    req.setId('11402cd5-4748-43b0-9868-cc0dd71244b3')
+
+    wardService.getWard(req, null, (err: grpcWeb.RpcError, res: GetWardResponse) => {
+      if (err) {
+        console.error('call failed', err)
+        return
+      }
+      console.log('call successful', res.getId(), res.getName())
+    })
+  }
```
JannikWibker, line 30: can you integrate this into the existing fetch+typechecking+mocking code? (namely [this](https://github.com/helpwave/web/blob/main/app/utils/api.ts); it should behave similar to [this here](https://github.com/helpwave/web/blob/main/app/utils/login.ts#L49-L57))

app/generated/Ward-svcServiceClientPb.ts
```diff
@@ -0,0 +1,172 @@
...
+
+// Code generated by protoc-gen-grpc-web. DO NOT EDIT.
+// versions:
+// 	protoc-gen-grpc-web v1.4.2
+// 	protoc              v3.20.3
+// source: ward-svc.proto
+
+
+/* eslint-disable */
+// @ts-nocheck
```
JannikWibker, line 15: this doesn't sound confidence inspiring
the generated code should at least be void of type errors

app/pages/index.tsx
JannikWibker: if that is not possible at least integration with `react-query` or `swr`
(the api of the generated code looks atrocious; some kind of nice wrapper around it is needed; what this wrapper will look like is up for debate / depends on how well the generated code can be encapsulated)

MaxSchaefer: It is only an example. Beside of the callback and error handling, IMO the API its not that _atrocious_. If we want to have a _"fetch-typechecking+mocking"_ wrapper (and yes, we should), we need to handle [gRPC Server-Side-Streaming](https://github.com/grpc/grpc-web#server-side-streaming) as well.

How would you implement such a wrapper? And should the implementation take place in the scope of this PR?

app/generated/Ward-svcServiceClientPb.ts
use-to: Disabling the typechecking does not imply errors inside the generated code.

One could even argue that generated source code is by definition (as you said) syntactically error-free and therefore does not need to be tested. You are welcome to choose how deep we should go in this discussion, or if we should dwell on it.

JannikWibker: disabling error checking affects the behaviour

generated code is definitely not always error free and using `@ts-nocheck` to circumvent checks implies that it is indeed not free of errors as it wouldn't be needed otherwise

MaxSchaefer: From a functional point of view, we should be satisfied with the codegen. If we do not like the code style, we should address this in a new issue after the merge.

app/pages/index.tsx
MaxSchaefer: I've adopted to promises instead of the callbacks. We should merge this example and move on as explained above.

use-to (approved):

app/pages/index.tsx
MaxSchaefer: #88

app/generated/Ward-svcServiceClientPb.ts
MaxSchaefer: https://github.com/helpwave/web/issues/89

PR #87, opened 2023-03-04
add PillLabels

### LINKED ISSUE #72: Add Pill / Label component

DasProffi:
closes #72
Unbenannt

### DIFF
--- a/app/components/PillLabel.tsx
+++ b/app/components/PillLabel.tsx
@@ -1,19 +1,21 @@
 import { tw } from '@helpwave/common/twind/index'
+import { TaskState } from '../enums/TaskState'
 
 export type PillLabelProps = {
   count?: number,
-  state?: 'unscheduled' | 'inProgress' | 'done'
+  state?: TaskState
 }
 
-const PillLabel = ({ count = 0, state = 'unscheduled' }: PillLabelProps) => {
-  // case 'unscheduled' is default
-  let labelNumber = '1'
-  let text = 'unscheduled'
-  switch (state) {
-    case 'inProgress': labelNumber = '2'; text = 'in progress'; break
-    case 'done': labelNumber = '3'; text = 'done'; break
-    default: break
+const PillLabel = ({ count = 0, state = TaskState.unscheduled }: PillLabelProps) => {
+  const textMap = {
+    [TaskState.unscheduled]: 'unscheduled',
+    [TaskState.inProgress]: 'in progress',
+    [TaskState.done]: 'done'
   }
+  const labelMap = { [TaskState.unscheduled]: '1', [TaskState.inProgress]: '2', [TaskState.done]: '3' }
+
+  const labelNumber = labelMap[state]
+  const text = textMap[state]
   // TODO Translation
   return (
     <div className={tw(`flex flex-row w-2/3 px-2 rounded-md justify-between bg-hw-label-${labelNumber}-background`)}>
--- a/app/components/PillLabel.tsx
+++ b/app/components/PillLabel.tsx
@@ -1,12 +1,13 @@
 import { tw } from '@helpwave/common/twind/index'
-import { TaskStates } from '../dataclasses/TaskState'
+import type { TaskStateInformation } from '../dataclasses/TaskState'
+import { TaskState } from '../dataclasses/TaskState'
 
 export type PillLabelProps = {
   count?: number,
-  state?: {text: string, label: number}
+  state?: TaskStateInformation
 }
 
-const PillLabel = ({ count = 0, state = TaskStates.unscheduled }: PillLabelProps) => {
+const PillLabel = ({ count = 0, state = TaskState.unscheduled }: PillLabelProps) => {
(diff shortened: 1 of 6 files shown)


MaxSchaefer (approved):
@DasProffi looks good! If you want, you can provide the implementation for the other representation in the issue as well.

app/components/PillLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const PillLabel = ({ count = 0, state = 'unscheduled' }: PillLabelProps) => {
+  // case 'unscheduled' is default
+  let labelNumber = '1'
+  let text = 'unscheduled'
+  switch (state) {
+    case 'inProgress': labelNumber = '2'; text = 'in progress'; break
+    case 'done': labelNumber = '3'; text = 'done'; break
+    default: break
+  }
```
use-to, line 16: use an map data structure instead
`switch` can contain any type of logic -a `map` is more restrictive

app/components/PillLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+  // TODO Translation
+  return (
+    <div className={tw(`flex flex-row w-2/3 px-2 rounded-md justify-between bg-hw-label-${labelNumber}-background`)}>
+      <div className={tw(`flex flex-row items-center text-hw-label-${labelNumber}-text`)}>
+        <div className={tw(`rounded-full w-2 h-2 bg-hw-label-${labelNumber}-accent`)} />
+        <div className={tw('w-2')} />
+        {text}
```
use-to, line 23: Is it the right font? Is `Inter` the default one?
There is the twind class `font-sans` but I'm not sure whether we have to use it. @JannikWibker ?

use-to (changes requested):

app/components/PillLabel.tsx
MaxSchaefer: The state can be an enum and should then be used in a map with the displayable text.

app/enums/TaskState.tsx
```diff
@@ -0,0 +1,7 @@
+enum TaskState {
+  unscheduled,
+  inProgress,
+  done
+}
+
+export { TaskState }
```
use-to, line 7: Why should we use a enum here?

```suggestion
export const TaskState = {
	unscheduled: {text: "unscheduled", label: 1},
	inProgress: {text: "in progress", label: 2},
	done: {text: "done", label: 3},
}
```

So just use:
```ts
const doneState = TaskState.done;
console.log(doneState.text, doneState.label)
```

use-to (changes requested):

app/components/PillLabel.tsx
use-to: @DasProffi just ensure using `Inter` here. We don't have to wait for @JannikWibker with this. It's more a general question of mine.

use-to:
> @DasProffi looks good! If you want, you can provide the implementation for the other representation in the issue as well.

I would skip that for the moment! Just concentrate on other issues/components.
Maybe create a new issues with low priority for that.

app/enums/TaskState.tsx
MaxSchaefer: Is also possible.

app/components/PillLabel.tsx
DasProffi: As far as I know, `inter` is the default for tailwind, therefore no additional tag should be required

app/components/PillLabel.tsx
```diff
@@ -0,0 +1,23 @@
+import { tw } from '@helpwave/common/twind/index'
+import { TaskStates } from '../dataclasses/TaskState'
+
+export type PillLabelProps = {
+  count?: number,
+  state?: {text: string, label: number}
```
use-to, line 6: create a type definition for this

app/dataclasses/TaskState.tsx
```diff
@@ -0,0 +1,5 @@
+export const TaskStates = {
+  unscheduled: { text: 'unscheduled', label: 1 },
+  inProgress: { text: 'in progress', label: 2 },
+  done: { text: 'done', label: 3 },
```
use-to, line 4: Add type definitions

app/dataclasses/TaskState.tsx
```diff
@@ -0,0 +1,5 @@
+export const TaskStates = {
```
use-to, line 1:
```suggestion
export const TaskState = {
```

use-to (changes requested):

app/dataclasses/TaskState.tsx
```diff
@@ -0,0 +1,7 @@
+export const TaskState: { [key in 'unscheduled' | 'inProgress' | 'done']: TaskStateInformation } = {
```
DasProffi, line 1: should the key only be a string?

use-to (approved):

use-to (approved):

PR #91, opened 2023-03-05
build(app): docker setup

MaxSchaefer:


Dockerfile.app
```diff
@@ -0,0 +1,37 @@
...
+COPY lib/package*.json ./lib/
+
+# install depedencies
+RUN pnpm install --frozen-lockfile
+
+# build app
+COPY . .
+RUN pnpm --filter $WS run build
+
+FROM node:18.12.0-alpine
```
use-to, line 23:
```suggestion
FROM node:18.12.0-alpine

LABEL maintainer="tech@helpwave.de"

```

Dockerfile.app
```diff
@@ -0,0 +1,37 @@
...
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 nextjs
+
+# copy transpiled app
+COPY --from=build --chown=nextjs:nodejs /web/app/build/standalone ./
+COPY --from=build --chown=nextjs:nodejs /web/app/public ./app/public
+COPY --from=build --chown=nextjs:nodejs /web/app/build/static ./app/build/static
+
+EXPOSE 3000
+ENV PORT 3000
```
use-to, line 36: This should be `80` as we serve a http server here (known ports).
Maybe use `ARG PORT=80` to be able to overwrite the port manually.
```suggestion
EXPOSE 80
ENV PORT 80
```

use-to (approved):

PR #92, opened 2023-03-07
ci(app): specify more branches and paths

MaxSchaefer:
I am not sure if this is working on main. Let's see.


use-to:
Why do we even scope this action?

MaxSchaefer:
> Why do we even scope this action?

This is a monorepo. Do we really want to rebuild the images on every commit?

use-to (approved):

use-to:
You're damn right!

PR #111, opened 2023-03-20
add PillLabelBox

### LINKED ISSUE #104: Ward tile
Ward Card_active

DasProffi:
closes #104

### DIFF
--- a/app/components/PillLabelBox.tsx
+++ b/app/components/PillLabelBox.tsx
@@ -1,0 +1,62 @@
+import { tw } from '@helpwave/common/twind/index'
+
+export type PillLabelBoxProps = {
+  unscheduledTasks: number,
+  inProgressTasks: number,
+  doneTasks: number
+}
+
+const PillLabelBox = ({ unscheduledTasks, inProgressTasks, doneTasks }: PillLabelBoxProps) => {
+  const between = '0.2rem'
+  const height = '0.75rem' // equivalent to h-6 of tailwind
+  return (
+    <div className={tw('flex flex-row h-6')}>
+      <div className={tw('flex flex-row bg-hw-label-1-background rounded-l-md pl-2 pr-1 items-center text-hw-label-1-text')}>
+        <div className={tw(`rounded-full w-2 h-2 bg-hw-label-1-accent mr-1`)}/>
+        {unscheduledTasks.toString()}
+      </div>
+      <div
+        className={tw('w-0 h-0 border-hw-label-1-background border-solid border-r-transparent border-b-transparent')}
+        style={{
+          borderBottomWidth: height,
+          borderRightWidth: between,
+          borderLeftWidth: between,
+          borderTopWidth: height
+        }}/>
+      <div
+        className={tw('w-0 h-0 border-hw-label-2-background border-solid border-l-transparent border-t-transparent')}
+        style={{
+          borderBottomWidth: height,
+          borderRightWidth: between,
+          borderLeftWidth: between,
+          borderTopWidth: height
+        }}/>
+      <div className={tw('flex flex-row bg-hw-label-2-background px-1 items-center text-hw-label-2-text')}>
+        <div className={tw(`rounded-full w-2 h-2 bg-hw-label-2-accent mr-1`)}/>
+        {inProgressTasks.toString()}
+      </div>
+      <div
+        className={tw('w-0 h-0 border-hw-label-2-background border-solid border-r-transparent border-b-transparent')}
+        style={{
+          borderBottomWidth: height,
+          borderRightWidth: between,
+          borderLeftWidth: between,
+          borderTopWidth: height
+        }}/>
+      <div
+        className={tw('w-0 h-0 border-hw-label-3-background border-solid border-l-transparent border-t-transparent')}
+        style={{
+          borderBottomWidth: height,
+          borderRightWidth: between,
+          borderLeftWidth: between,
+          borderTopWidth: height
+        }}/>
+      <div className={tw('flex flex-row bg-hw-label-3-background rounded-r-md pl-1 pr-2 items-center text-hw-label-3-text')}>
+        <div className={tw(`rounded-full w-2 h-2 bg-hw-label-3-accent mr-1`)}/>
+        {doneTasks.toString()}
+      </div>
+    </div>
+  )
+}
+
+export { PillLabelBox }
--- a/app/components/WardTile.tsx
+++ b/app/components/WardTile.tsx
@@ -1,0 +1,61 @@
+import { tw } from '@helpwave/common/twind/index'
+import type { Languages } from '@helpwave/common/hooks/useLanguage'
+import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
+import { useTranslation } from '@helpwave/common/hooks/useTranslation'
+import type { FunctionComponent } from 'react'
+import Bed from '../icons/Bed'
+import { PillLabelBox } from './PillLabelBox'
+
+type WardTileTranslation = {
+  edit: string
+}
+
+const defaultWardTileTranslations: Record<Languages, WardTileTranslation> = {
+  en: {
+    edit: 'Edit'
+  },
+  de: {
+    edit: 'Bearbeiten'
+  }
+}
+
+export type WardTileProps = {
+  wardName: string,
+  roomNames: string[],
+  bedCount: number,
+  unscheduledTasks: number,
+  inProgressTasks: number,
+  doneTasks: number,
+  isSelected: boolean,
+  tileClickedCallback: () => void,
+  editClickedCallback: () => void
+}
+
+const WardTile: FunctionComponent<PropsWithLanguage<WardTileTranslation, WardTileProps>> = (props) => {
+  const translation = useTranslation(props.language, defaultWardTileTranslations)
+  // TODO change border color for onHover and selected
+  const selectedBorderColor = props.isSelected ? 'border-black' : ''
+  return (
+    <div onClick={props.tileClickedCallback}
+         className={tw(`${selectedBorderColor} group cursor-pointer rounded-md py-2 px-4 border hover:border-black w-full`)}>
+      <div className={tw('flex flex-row justify-between w-full')}>
+        <h5 className={tw('font-bold')}>{props.wardName}</h5>
+        <button onClick={props.editClickedCallback}
+                className={tw('hidden group-hover:block')}>{translation.edit}</button>
+      </div>
+      <div className={tw('text-left my-1')}>{props.roomNames.join(', ')}</div>
+      <div className={tw('flex flex-row justify-between w-full')}>
+        <div className={tw('flex flex-row')}>
+          <Bed/>
+          <div className={tw('pl-1')}>{props.bedCount}</div>
+        </div>
+        <PillLabelBox
+          unscheduledTasks={props.unscheduledTasks}
+          inProgressTasks={props.inProgressTasks}
+          doneTasks={props.doneTasks}/>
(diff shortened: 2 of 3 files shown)


app/components/PillLabelBox.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const height = '0.75rem' // equivalent to h-6 of tailwind
+  return (
+    <div className={tw('flex flex-row h-6')}>
+      <div className={tw('flex flex-row bg-hw-label-1-background rounded-l-md pl-2 pr-1 items-center text-hw-label-1-text')}>
+        <div className={tw(`rounded-full w-2 h-2 bg-hw-label-1-accent mr-1`)}/>
+        {unscheduledTasks.toString()}
+      </div>
+      <div
+        className={tw('w-0 h-0 border-hw-label-1-background border-solid border-r-transparent border-b-transparent')}
+        style={{
```
MaxSchaefer, line 20: [You can use arbitrary values in tailwind.](https://tailwindcss.com/docs/border-width#arbitrary-values) We can adopt this here like `className={tw('... border-y-[0.75rem] border-x-[0.2rem]')}` instead of the inline styles.

app/components/WardTile.tsx
```diff
@@ -0,0 +1,61 @@
...
+const defaultWardTileTranslations: Record<Languages, WardTileTranslation> = {
+  en: {
+    edit: 'Edit'
+  },
+  de: {
+    edit: 'Bearbeiten'
+  }
+}
+
+export type WardTileProps = {
```
MaxSchaefer, line 22: We should definitely create a separate type for a ward and pass this newly created type beside the specific props for that component.

app/components/WardTile.tsx
```diff
@@ -0,0 +1,61 @@
...
+  wardName: string,
+  roomNames: string[],
+  bedCount: number,
+  unscheduledTasks: number,
+  inProgressTasks: number,
+  doneTasks: number,
+  isSelected: boolean,
+  tileClickedCallback: () => void,
```
MaxSchaefer, line 30: Default values for callbacks? Like `() => {}`?

app/components/PillLabelBox.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export type PillLabelBoxProps = {
+  unscheduledTasks: number,
+  inProgressTasks: number,
+  doneTasks: number
+}
+
+const PillLabelBox = ({ unscheduledTasks, inProgressTasks, doneTasks }: PillLabelBoxProps) => {
+  const between = '0.2rem'
+  const height = '0.75rem' // equivalent to h-6 of tailwind
```
MaxSchaefer, line 11: [`h-6: height: 1.5rem; /* 24px */`](https://tailwindcss.com/docs/height#class-reference)

app/components/PillLabelBox.tsx
```diff
@@ -0,0 +1,62 @@
+import { tw } from '@helpwave/common/twind/index'
+
+export type PillLabelBoxProps = {
+  unscheduledTasks: number,
```
MaxSchaefer, line 4: We should remove the suffix task. In the future, we should have a separate type called "task". Can be confusing.

MaxSchaefer (changes requested):

MaxSchaefer:
Looks nice 😍
Bildschirm­foto 2023-03-20 um 8 48 04 PM

app/components/WardTile.tsx
DasProffi: You cannot set a default value on a type definition as far as I know. Do you want me to a default Object like
```
const default = {
    tileClickedCallback: () => {}
}
```

app/components/PillLabelBox.tsx
DasProffi: Good to know, thank you!

app/components/WardTile.tsx
MaxSchaefer: Take a look at [this example](https://github.com/helpwave/web/blob/main/app/components/ProfilePicture.tsx#L9) for default values for props.

app/icons/Bed.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { forwardRef } from 'react'
+import type { SVGProps } from 'react'
+
+export default forwardRef<SVGSVGElement, SVGProps<SVGSVGElement>>(function Bed(props, ref) {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor"
```
JannikWibker, line 6: remove the `xmlsns="..."`; it isn't needed
also `props` and `ref` are currently unused, check out https://github.com/helpwave/web/blob/main/app/icons/Check.tsx maybe

app/components/WardTile.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ward,
+  tileClickedCallback = () => undefined,
+  editClickedCallback = () => undefined
+}: PropsWithLanguage<WardTileTranslation, WardTileProps>) => {
+  const translation = useTranslation(language, defaultWardTileTranslations)
+  // TODO change border color for onHover and selected
+  const selectedBorderColor = isSelected ? 'border-black' : ''
+  return (
+    <div onClick={tileClickedCallback}
+         className={tw(`${selectedBorderColor} group cursor-pointer rounded-md py-2 px-4 border hover:border-black w-full`)}>
```
JannikWibker, line 50: you can also do the following: `tx('group cursor-pointer ...', { 'border-black': isSelected })`

app/components/WardTile.tsx
```diff
@@ -0,0 +1,71 @@
...
+  unscheduledTasks: number,
+  inProgressTasks: number,
+  doneTasks: number
+}
+
+export type WardTileProps = {
+  ward: WardDTO,
+  isSelected: boolean,
+  tileClickedCallback?: () => void,
+  editClickedCallback?: () => void
```
JannikWibker, line 35: `onTileClick` and `onEditClick` instead

app/components/WardTile.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+type WardDTO = {
+  name: string,
+  roomNames: string[],
+  bedCount: number,
+  unscheduledTasks: number,
+  inProgressTasks: number,
+  doneTasks: number
```
JannikWibker, line 28: I would prefer `tasks: { unscheduled: number, inProgress: number, done: number }` a bit more

app/components/WardTile.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export type WardTileProps = {
+  ward: WardDTO,
+  isSelected: boolean,
+  tileClickedCallback?: () => void,
+  editClickedCallback?: () => void
+}
+
+const WardTile: FunctionComponent<PropsWithLanguage<WardTileTranslation, WardTileProps>> = ({
```
JannikWibker, line 38: remove `FunctionComponent<PropsWithLanguage<WardTileTranslation, WardTileProps>>`, you already have a type annotation for the props object, no need to specify this twice (the props object is the preferred way as react has some odd types sometimes, thus it is best to rely on inference to do the heavy lifting)

app/components/WardTile.tsx
```diff
@@ -0,0 +1,71 @@
+import { tw } from '@helpwave/common/twind/index'
+import type { Languages } from '@helpwave/common/hooks/useLanguage'
+import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
+import { useTranslation } from '@helpwave/common/hooks/useTranslation'
+import type { FunctionComponent } from 'react'
+import Bed from '../icons/Bed'
```
JannikWibker, line 6: could probably move this to `/lib`, isn't required though

app/components/PillLabelBox.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export type PillLabelBoxProps = {
+  unscheduledCount: number,
+  inProgressCount: number,
+  doneCount: number
+}
+
+const PillLabelBox = ({ unscheduledCount, inProgressCount, doneCount }: PillLabelBoxProps) => {
+  const between = '0.2rem'
+  const height = '0.75rem' // equivalent to 1.5rem / 2 = 24px / 2 = h-6 / 2 of tailwind, half because border is set on both sides
```
JannikWibker, line 11: it's better to use pixels imo

app/components/PillLabelBox.tsx
```diff
@@ -0,0 +1,45 @@
+import { tw } from '@helpwave/common/twind/index'
+
+export type PillLabelBoxProps = {
+  unscheduledCount: number,
+  inProgressCount: number,
+  doneCount: number
```
JannikWibker, line 6: the `Count` suffix is pretty much implied by context (namely the type `number` and the fact that there's 3 props)

app/components/PillLabelBox.tsx
```diff
@@ -0,0 +1,45 @@
...
+const PillLabelBox = ({ unscheduledCount, inProgressCount, doneCount }: PillLabelBoxProps) => {
+  const between = '0.2rem'
+  const height = '0.75rem' // equivalent to 1.5rem / 2 = 24px / 2 = h-6 / 2 of tailwind, half because border is set on both sides
+  const borderWidth = `border-y-[${height}] border-x-[${between}]`
+  return (
+    <div className={tw('flex flex-row h-6')}>
+      <div
+        className={tw('flex flex-row bg-hw-label-1-background rounded-l-md pl-2 pr-1 items-center text-hw-label-1-text')}>
+        <div className={tw(`rounded-full w-2 h-2 bg-hw-label-1-accent mr-1`)}/>
+        {unscheduledCount.toString()}
```
JannikWibker, line 18: no need for `.toString()`

app/components/WardTile.tsx
```diff
@@ -0,0 +1,71 @@
...
+const defaultWardTileTranslations: Record<Languages, WardTileTranslation> = {
+  en: {
+    edit: 'Edit'
+  },
+  de: {
+    edit: 'Bearbeiten'
+  }
+}
+
+type WardDTO = {
```
JannikWibker, line 22: what does `DTO` stand for?

app/components/WardTile.tsx
JannikWibker, line 50: the border should be a bit thicker
the hover color should be `hw-primary-700` (the colors for 700 and 800 seem to be messed up a bit in figma; use the lighter one of the two; I'll fix the figma issue)

app/components/WardTile.tsx
DasProffi: data transfer object, it's only there for now to simulate a ward object which we should introduce in some other issue

app/components/WardTile.tsx
```diff
@@ -0,0 +1,71 @@
...
+  editClickedCallback = () => undefined
+}: PropsWithLanguage<WardTileTranslation, WardTileProps>) => {
+  const translation = useTranslation(language, defaultWardTileTranslations)
+  // TODO change border color for onHover and selected
+  const selectedBorderColor = isSelected ? 'border-black' : ''
+  return (
+    <div onClick={tileClickedCallback}
+         className={tw(`${selectedBorderColor} group cursor-pointer rounded-md py-2 px-4 border hover:border-black w-full`)}>
+      <div className={tw('flex flex-row justify-between w-full')}>
+        <h5 className={tw('font-bold')}>{ward.name}</h5>
```
JannikWibker, line 52: use a `span` instead and add the necessary styles (`font-space font-bold` should do)

MaxSchaefer (approved):

PR #114, opened 2023-03-21
chore: rename app folder to tasks

### LINKED ISSUE #113: Rename app folder to tasks

florian-sabonchi:
Closes #113


tasks/components/Input.tsx
MaxSchaefer, line 1: @JannikWibker can you take a look 👀?

MaxSchaefer:
[Can you take a look at this commit in particular?](https://github.com/helpwave/web/pull/114/commits/5e12a3beb38f6c7e136c60753cba90ed3d91bc13)

Dockerfile.app
```diff
@@ -7,16 +7,16 @@ RUN curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm
...
-COPY app/package*.json ./app/
+COPY tasks/package*.json ./tasks/
 COPY lib/package*.json ./lib/
 
-# install depedencies
+# install dependencies
 RUN pnpm install --frozen-lockfile
 
-# build app
+# build tasks
```
JannikWibker, line 19: `# build tasks app` sounds a bit better

tasks/components/Input.tsx
```diff
@@ -22,7 +22,7 @@ type InputProps = {
    * @default noop
    */
   onChange?: (text: string) => void
-} & Omit<InputHTMLAttributes<Element>, 'id' | 'value' | 'label' | 'type' | 'onChange'>
+} & Omit<InputHTMLAttributes<HTMLInputElement>, 'id' | 'value' | 'label' | 'type' | 'onChange' | 'crossOrigin'>
```
JannikWibker, line 25: I'm not sure why `'crossOrigin'` was added here?

JannikWibker:
Remove the `.lintstagedrc.json` file, I removed husky a while back and forgot about that file

Dockerfile.app
JannikWibker: As a side note: why is this file called `Dockerfile.app`, shouldn't this just be `Dockerfile` and live in the `/app` directory?

florian-sabonchi: Yes i agree

tasks/components/Input.tsx
florian-sabonchi: I am also not sure

JannikWibker: I'll investigate the CI / Docker issue (later today)

JannikWibker: The docker setup seems to be a bit messy right now, pretty much all your changes look good but I might need to do some clean up when it comes to docker in order to resolve this

Dockerfile.app
MaxSchaefer: Due to the pnpm workspaces.

tasks/components/Input.tsx
MaxSchaefer: Can you list your change requests, then someone else can take them over if necessary. Please open a new issue, not in the scope of this PR.

MaxSchaefer (approved):

Dockerfile.app
JannikWibker, line 1: Rename this file to `tasks/Dockerfile`

This requires an update to the corresponding github action as well

With this change you can now build docker images locally without having to do `docker build - < cat Dockerfile.app`

Dockerfile.app
```diff
@@ -7,16 +7,16 @@ RUN curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm
 
 WORKDIR /web
 
-# copy all dependency declerations inside the container
+# copy all dependency declarations inside the container
 COPY package*.json .
 COPY pnpm*.yaml .
```
JannikWibker, line 12: Don't try to be smart here, just list the files individually; its not like there is an enormous amount of them

Or copy all files (for locally building the containers I suggest adding a `.dockerignore` file to ignore `node_modules` etc.)

Dockerfile.app
```diff
@@ -7,16 +7,16 @@ RUN curl -f https://get.pnpm.io/v6.16.js | node - add --global pnpm
...
-COPY app/package*.json ./app/
+COPY tasks/package*.json ./tasks/
 COPY lib/package*.json ./lib/
 
-# install depedencies
+# install dependencies
 RUN pnpm install --frozen-lockfile
```
JannikWibker, line 17: do `CI=1 pnpm install` instead

JannikWibker, line 17: you need to copy the whole `lib/` folder before doing `pnpm install` as `lib/` is a dependency of `tasks/`

.lintstagedrc.json
JannikWibker, line 1: Remove this file

Dockerfile.app
MaxSchaefer: @JannikWibker @florian-sabonchi 

No need to redirect `cat Dockerfile.tasks` into `docker build`, there is a flag for that. You can use `docker build --file Dockerfile.tasks .`.

---

#### Now, some explanation why our Dockerfiles should not be inside the corresponding src directory.

In every case, we need to run the `docker build` command from the repository root context. It is not possible to copy files into the container that are not inside the context directory. Something like `COPY ../lib ./lib` is not possible.

From [Dockerfile reference](https://docs.docker.com/engine/reference/builder/#add):
> The <src> path must be inside the context of the build; you cannot ADD ../something /something, because the first step of a docker build is to send the context directory (and subdirectories) to the docker daemon.

That means, the `COPY` statements, are always from the context root. This results in the following confusing example:

`tasks/Dockerfile`
```Dockerfile
...
COPY lib/package.json ./lib # The Dockerfile lives inside tasks/Dockerfile but gets called from the context root.
COPY package.json ./tasks # Will not copy tasks/package.json, but /package.json from the context root.
...
```

To get rid of this confusing setup/dx, we can just leave the Dockerfile in the context root. This results in the following build command:
```shell
docker build --file Dockerfile.tasks .
```

*fyi: We already adopted this behavior in [helpwave/services](https://github.com/helpwave/services).*

PR #115, opened 2023-03-21
add task template column

### LINKED ISSUE #107: Task template list column
Bildschirm­foto 2023-03-19 um 4 36 27 PM

DasProffi:
closes #107
Unbenannt

### DIFF
--- a/app/components/TaskTemplateListColumn.tsx
+++ b/app/components/TaskTemplateListColumn.tsx
@@ -1,0 +1,66 @@
+import { tw } from '@helpwave/common/twind/index'
+import type { Languages } from '@helpwave/common/hooks/useLanguage'
+import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
+import { useTranslation } from '@helpwave/common/hooks/useTranslation'
+import { useState } from 'react'
+import { TaskTemplateTile } from './TaskTemplateTile'
+import Add from '../icons/Add'
+
+export type TaskTemplateListColumnTranslation = {
+  name: string,
+  addNewTaskTemplate: string
+}
+
+type TaskTemplateDTO = {
+  name: string,
+  subtaskCount: number
+}
+
+export type TaskTemplateListColumnProps = {
+  translations: Record<Languages, TaskTemplateListColumnTranslation>,
+  taskTemplates: TaskTemplateDTO[]
+}
+
+export const TaskTemplateListColumn = ({
+  language,
+  translations,
+  taskTemplates
+}: PropsWithLanguage<TaskTemplateListColumnTranslation, TaskTemplateListColumnProps>) => {
+  const [selected, setSelected] = useState<TaskTemplateDTO | undefined>(undefined)
+  const translation = useTranslation(language, translations)
+
+  // TODO update with opening of the create template screen
+  const addNewTask = () => undefined
+
+  const onTileClicked = (taskTemplate: TaskTemplateDTO) => {
+    // TODO open template edit screen
+    setSelected(taskTemplate)
+  }
+
+  // TODO action on clicking the template tile
+  const onEditClicked = (taskTemplate: TaskTemplateDTO) => taskTemplate
+
+  return (
+    <div>
+      <div className={tw('flex flex-row items-center')}>
+        <div className={tw('w-2 h-2 mx-2 rounded-full bg-gray-300')}/>
+        <span className={tw('font-bold')}>{translation.name}</span>
+      </div>
+      {taskTemplates.map(taskTemplate => (
+          <div key={taskTemplate.name} className={tw('my-2')}>
+            <TaskTemplateTile name={taskTemplate.name}
+                              subtaskCount={taskTemplate.subtaskCount}
+                              isSelected={selected === taskTemplate}
+                              onEditClicked={() => onEditClicked(taskTemplate)}
+                              onTileClicked={() => onTileClicked(taskTemplate)}/>
+          </div>
+      )
+      )}
+      <div onClick={addNewTask}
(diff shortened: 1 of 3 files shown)


app/components/TaskTemplateListColumn.tsx
```diff
@@ -0,0 +1,66 @@
...
+  addNewTaskTemplate: string
+}
+
+type TaskTemplateDTO = {
+  name: string,
+  subtaskCount: number
+}
+
+export type TaskTemplateListColumnProps = {
+  translations: Record<Languages, TaskTemplateListColumnTranslation>,
```
MaxSchaefer, line 20: Why do you propagate the translations back to the caller? Is there any reason why not handle the translations the same as in https://github.com/helpwave/web/blob/issue/116-Stacked_Profile_Pictures_Component/app/components/StackedProfiles.tsx for example?

app/icons/Add.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { forwardRef } from 'react'
+import type { SVGProps } from 'react'
+
+export default forwardRef<SVGSVGElement, SVGProps<SVGSVGElement>>(function Bed(props, ref) {
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor"
```
MaxSchaefer, line 6: `xmlns` seams to be unnecessary here.

MaxSchaefer (changes requested):

app/components/TaskTemplateListColumn.tsx
DasProffi: Because the translation is based upon the name of the column for which there are at least 2 different cases. Allowing the caller to set the translation allows an easy addition of another column.

But I will make a case distinction for both possible column names, since the likelyhood of an additional one in the near future is low and not worth violating the pattern

app/components/TaskTemplateListColumn.tsx
```diff
@@ -0,0 +1,76 @@
+import { tw } from '@helpwave/common/twind/index'
+import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
+import { useTranslation } from '@helpwave/common/hooks/useTranslation'
+import { useState } from 'react'
+import { TaskTemplateTile } from './TaskTemplateTile'
+import Add from '../icons/Add'
```
JannikWibker, line 6: move this to `/lib`, this will be used quite a lot probably

app/components/TaskTemplateListColumn.tsx
```diff
@@ -0,0 +1,76 @@
+import { tw } from '@helpwave/common/twind/index'
+import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
+import { useTranslation } from '@helpwave/common/hooks/useTranslation'
+import { useState } from 'react'
```
JannikWibker, line 4: nit: order imports by "importance", meaning react goes on top, then next, then external libraries, then own imports, icons come last

app/components/TaskTemplateListColumn.tsx
```diff
@@ -0,0 +1,76 @@
...
+                              subtaskCount={taskTemplate.subtaskCount}
+                              isSelected={selected === taskTemplate}
+                              onEditClicked={() => onEditClicked(taskTemplate)}
+                              onTileClicked={() => onTileClicked(taskTemplate)}/>
+          </div>
+      )
+      )}
+      <div onClick={addNewTask}
+           className={tw('flex flex-row w-full h-16 rounded-md border-2 hover:border-hw-primary-700 items-center justify-center cursor-pointer mt-4')}>
+        <Add/>
```
JannikWibker, line 71: `<Add />` with a space

app/icons/Add.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { forwardRef } from 'react'
+import type { SVGProps } from 'react'
+
+export default forwardRef<SVGSVGElement, SVGProps<SVGSVGElement>>(function Bed(props, ref) {
+  return (
+    <svg width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor"
+         strokeWidth="2" strokeLinecap="round" strokeLinejoin="round" ref={ref} {...props}>
```
JannikWibker, line 7: Similar formatting as described above, for icons this isn't really all that important though

app/components/TaskTemplateListColumn.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <div className={tw('w-2 h-2 mx-2 rounded-full bg-gray-300')}/>
+        <span className={tw('font-bold')}>{translation.columTitle(isWardTemplateColumn)}</span>
+      </div>
+      {taskTemplates.map(taskTemplate => (
+          <div key={taskTemplate.name} className={tw('my-2')}>
+            <TaskTemplateTile name={taskTemplate.name}
+                              subtaskCount={taskTemplate.subtaskCount}
+                              isSelected={selected === taskTemplate}
+                              onEditClicked={() => onEditClicked(taskTemplate)}
+                              onTileClicked={() => onTileClicked(taskTemplate)}/>
```
JannikWibker, line 65: `onTileClick={setSelected}` and
`onEditClick={onEditClick}` (assuming the actual implementation isn't going to be short enough to inline here)

DasProffi: this way you would have to pass down the actual object just to receive it again, while this way we only pass the relevant attributes and remember the template

DasProffi: These functions need to be extended later on, because a click would open the update screen in the right column

MaxSchaefer (approved):

JannikWibker (approved):

PR #117, opened 2023-03-21
add Stacked Profiles

### LINKED ISSUE #116: Stacked Profile Pictures Component
As mentioned in #100 the stacked profile pictures will be implemented through this issue

DasProffi:
closes #116


florian-sabonchi (approved):

app/components/StackedProfiles.tsx
```diff
@@ -0,0 +1,51 @@
...
+  en: {
+    others: (users: number) => users === 1 ? 'other' : 'others'
+  },
+  de: {
+    others: (users: number) => users === 1 ? 'weiterer' : 'weitere'
+  }
+}
+
+// TODO replace later
+type UserDTO = {
```
MaxSchaefer, line 20: I think the color should not be a part of the user. We should inline these styles into `StackedProfilesProps` instead.

app/components/StackedProfiles.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const height = 24 // 24px
+  const stackingOverlap = 0.5 // given as a percentage
+
+  return (
+    <div className={tw(`h-[${height + 'px'}] flex flex-row relative`)}>
+      {displayedProfiles.map((user, index) => (
+        <div key={user.name}
+             className={tx(`h-[${height + 'px'}] w-[${height + 'px'}] rounded-full border-1 bg-[${user.color}] absolute left-[${(index * height * stackingOverlap) + 'px'}] z-[${maxShownProfiles - index}]`)}/>
+      ))}
+      <div className={tx(`ml-[${((maxShownProfiles + 1) * height * stackingOverlap + 4) + 'px'}]`, { hidden: users.length <= maxShownProfiles })}>
```
MaxSchaefer, line 46: Wrap an if-Statement around this div instead of `hidden` by using reacts conditional rendering.

app/components/StackedProfiles.tsx
```diff
@@ -0,0 +1,51 @@
...
+    others: (users: number) => users === 1 ? 'other' : 'others'
+  },
+  de: {
+    others: (users: number) => users === 1 ? 'weiterer' : 'weitere'
+  }
+}
+
+// TODO replace later
+type UserDTO = {
+  color: string,
```
MaxSchaefer, line 21: @JannikWibker is there a type for this?

MaxSchaefer (changes requested):

MaxSchaefer:
Also, we don't have borders around the colored circles in Figma, do we?

app/components/StackedProfiles.tsx
DasProffi: There should be no color property on the user later on, only a link to profile picture as far as I know.

We should finalize the DataModels to make it easer on implementation

app/components/StackedProfiles.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const stackingOverlap = 0.5 // given as a percentage
+
+  return (
+    <div className={tw(`h-[${height + 'px'}] flex flex-row relative`)}>
+      {displayedProfiles.map((user, index) => (
+        <div key={user.name}
+             className={tx(`h-[${height + 'px'}] w-[${height + 'px'}] rounded-full border-1 bg-[${user.color}] absolute left-[${(index * height * stackingOverlap) + 'px'}] z-[${maxShownProfiles - index}]`)}/>
+      ))}
+      <div className={tx(`ml-[${((maxShownProfiles + 1) * height * stackingOverlap + 4) + 'px'}]`, { hidden: users.length <= maxShownProfiles })}>
+        <span>{`+ ${(users.length - maxShownProfiles)} ${translation.others(users.length - maxShownProfiles)}`}</span>
```
use-to, line 47: `let x = user.length - maxShownProfiles`

app/components/StackedProfiles.tsx
```diff
@@ -0,0 +1,51 @@
...
+  others: (users: number) => string
+}
+
+const defaultStackedProfilesTranslations: Record<Languages, StackedProfilesTranslation> = {
+  en: {
+    others: (users: number) => users === 1 ? 'other' : 'others'
+  },
+  de: {
+    others: (users: number) => users === 1 ? 'weiterer' : 'weitere'
+  }
```
use-to, line 16: Switch to `others: (multipleUsers: boolean)`

use-to (changes requested):

DasProffi:
> Also, we don't have borders around the colored circles in Figma, do we?

There is a border arround them, see the following picture
Members (1)

app/components/StackedProfiles.tsx
MaxSchaefer: Okay, can you then add a `TODO` to replace the colors with profile picture urls?

DasProffi: Already updated it to use URL

app/components/StackedProfiles.tsx
JannikWibker, line 1: AvatarGroup is a better name

app/components/StackedProfiles.tsx
```diff
@@ -0,0 +1,58 @@
...
+  others: (isMultipleUsers: boolean) => string
+}
+
+const defaultStackedProfilesTranslations: Record<Languages, StackedProfilesTranslation> = {
+  en: {
+    others: (isMultipleUsers: boolean) => isMultipleUsers ? 'other' : 'others'
+  },
+  de: {
+    others: (isMultipleUsers: boolean) => isMultipleUsers ? 'weiterer' : 'weitere'
+  }
```
JannikWibker, line 17: this logic shouldn't be part of the translation object: `otherSingular`, `otherPlural` is fine as well

app/components/StackedProfiles.tsx
JannikWibker: Currently there is no such thing as a "hover effect" with a tooltip for the user, this is something that could exist in the future though

But until then we don't need to talk about what the "user object" needs to look like, the avatar/profile picture component can just get a url and be done with it, no need to pass more than that to the avatar/profile picture component.

app/components/StackedProfiles.tsx
DasProffi: we should be consistend we already have a profile picture, so we should name it `ProfileGroup` or the other one `Avatar` (or something similar)

JannikWibker: renaming the other component to `Avatar` was implied, could have worded that better / made that clearer

MaxSchaefer (changes requested):

MaxSchaefer (approved):

MaxSchaefer (approved):

JannikWibker:
@MaxSchaefer 

> Also, we don't have borders around the colored circles in Figma, do we?

I think we do, I added them iirc; if not then this should be updated in figma

JannikWibker (approved):

use-to (approved):

use-to:
Sorry 😅

PR #118, opened 2023-03-21
add organization member display

### LINKED ISSUE #102: Organization members list
Members Section

DasProffi:
closes #102
Unbenannt


app/icons/ChevronLast.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { forwardRef } from 'react'
```
JannikWibker, line 1: Similar as previous PRs: move this to `/lib`, it'll be used by multiple projects (applies to all icons pretty much)

app/components/PageIndicator.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <button onClick={() => changePage(0)} disabled={onFirstPage}>
+        <ChevronFirst className={tx({ 'opacity-30': onFirstPage })}/>
+      </button>
+      <button onClick={() => changePage(currentPage - 1)} disabled={onFirstPage}>
+        <ChevronLeft className={tx({ 'opacity-30': onFirstPage })}/>
+      </button>
+      <div className={tw('flex min-w-[80px] justify-center mx-2')}>
+        <span className={tw('select-none text-right flex-1')}>{currentPage + 1}</span>
+        <span className={tw('select-none mx-2')}>{translation.of}</span>
+        <span className={tw('select-none text-left flex-1')}>{numberOfPages}</span>
```
JannikWibker, line 56: You don't need 3 elements for this, you can just use a single one and do something like this:
```tsx
<span className={tw('select none')}>{translation.pageXofY(currentPage + 1, numberOfPages)}</span>
```

app/components/OrganizationMemberList.tsx
```diff
@@ -0,0 +1,161 @@
...
+  edit: string,
+  remove: string,
+  deselectAll: string,
+  selectAll: string,
+  members: string,
+  addMember: string,
+  saveChanges: string,
+  role: string,
+  of: string,
+  roleTypes: (role: Role) => string
```
JannikWibker, line 26: Functions are a nice thing to have in the translation objects but we shouldn't overuse them, they are pretty much just meant for things where you would otherwise use some kind of format string and insert values into it

In this case the following would be better:
```ts
export type OrganizationMemberListTranslation = {
  edit: string,
  // ...
  roles: Record<Role, string>
}
```

app/components/OrganizationMemberList.tsx
```diff
@@ -0,0 +1,161 @@
...
+import Dropdown from '../icons/TriangleDown'
+import { useState } from 'react'
+import { Checkbox } from './Checkbox'
+import { PageIndicator } from './PageIndicator'
+import { ProfilePicture } from './ProfilePicture'
+
+export enum Role {
+  user,
+  admin,
+}
```
JannikWibker, line 14: don't use enums in typescript, rather do something like this:
```ts
const roles = ['user', 'admin'] as const
export type Role = typeof roles
```

app/components/PageIndicator.tsx
JannikWibker, line 1: This should be called `Pagination.tsx` instead

app/components/OrganizationMemberList.tsx
```diff
@@ -0,0 +1,161 @@
...
+  name: string,
+  members: OrgMember[]
+}
+
+export type OrganizationMemberListProps = {
+  organization: OrganizationDTO,
+  usersPerPage?: number
+}
+
+export const OrganizationMemberList = ({
```
JannikWibker, line 77: I suggest using something like `@tanstack/table` (previously known as `react-table` instead, tables are hard to get right and without a headless component library aiding you in this endeavour it becomes even more difficult to get right

You could argue that it is overkill for such a small table (I would argue against that still) but just for the learning experience (there will be a lot more tables in the future most likely) it is worth to get started with it now

JannikWibker (commented):
I know this is just a draft PR right now but I still wanted to make some comments before too much time is spent doing things that would be noted in a later review regardless

app/components/PageIndicator.tsx
DasProffi: The issue isn't with the translation, but without three elements the numbers jump arround (the position changes) for different numbers which is annoying an distracts the user

DasProffi: but that is still annoying if the text shifts, therefore I solved that. Why would we now move back to a solution which is 2 lines shorter but worse

JannikWibker: Because having this split into multiple parts is overly complicated and doesn't allow all that much flexibility while translating; as an example this doesn't allow converting the numbers to another representation such as `"page one of two"`. This flexibility isn't necessarily needed but you also should make the code more complicated as it needs to be when you can just rely on the inline layouting to do the job. Having this be 3 dom elements might even seem overly complicated and over-engineered to users

app/components/OrganizationMemberList.tsx
```diff
@@ -0,0 +1,250 @@
...
+          </button>
+          <button onClick={saveChanges} disabled={disableSaveChanges} className={tx('px-2' + buttonStyle)}>{translation.saveChanges}</button>
+        </div>
+      </div>
+      <table
+        className={tw(`border-spacing-y-[8px] border-separate h-[${8 * (usersPerPage + 2) + rowHeight * usersPerPage + headerHeight}px]`)}>
+        <thead>
+        <tr className={tw(`h-[${headerHeight}px]`)}>
+          <th>
+            <div className={tw('flex flex-row justify-start')}>
```
JannikWibker, line 183: Don't overuse flex layouts, there are many places where it isn't needed at all

app/components/OrganizationMemberList.tsx
```diff
@@ -0,0 +1,250 @@
...
+        <span className={tw('font-bold font-space')}>{translation.members + ` (${members.length})`}</span>
+        <div className={tw('flex flex-row')}>
+          <button onClick={addUser} className={tw('flex flex-row items-center px-2 mr-2' + buttonStyle)}>
+            <span className={tw('mr-2')}>{translation.addMember}</span><Dropdown/>
+          </button>
+          <button onClick={saveChanges} disabled={disableSaveChanges} className={tx('px-2' + buttonStyle)}>{translation.saveChanges}</button>
+        </div>
+      </div>
+      <table
+        className={tw(`border-spacing-y-[8px] border-separate h-[${8 * (usersPerPage + 2) + rowHeight * usersPerPage + headerHeight}px]`)}>
```
JannikWibker, line 179: You shouldn't need to calculate the height at all, it is really bad practice to do something like this

Also prefer using the tailwind defaults (`border-spacing-y-2`) instead of custom pixel values

app/components/Pagination.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const onFirstPage = page === 0
+  const onLastPage = page === numberOfPages - 1
+
+  return (
+    <div className={tw('flex flex-row')}>
+      <button onClick={() => changePage(0)} disabled={onFirstPage}>
+        <ChevronFirst className={tx({ 'opacity-30': onFirstPage })}/>
+      </button>
+      <button onClick={() => changePage(page - 1)} disabled={onFirstPage}>
+        <ChevronLeft className={tx({ 'opacity-30': onFirstPage })}/>
```
JannikWibker, line 49: the icon uses `currentColor` internally, meaning you can recolor it by changing the text color. That is the preferred approach for this: `tw(onFirstPage ? 'text-...' : 'text-...')`

lib/twind/config.ts
```diff
@@ -42,23 +42,31 @@ export const config = defineConfig({
       colors: {
         // TODO: custom helpwave (hw) colors go here
```
JannikWibker, line 43: I used these same colors [here](https://github.com/JannikWibker/docs/blob/main/lib/twind/config.ts) for testing around a bit, copy them from there ideally.

The advantage of that is that all the colors are neatly separated into 3 groups: base colors, light colors and dark colors

app/components/PageIndicator.tsx
DasProffi: I would evaluate not jumping text higher than reducing the number of dom elements by two, if the solution is still readable without much effort, which it currently is

If we later on encounter a situation where a single text would be better, we can still change it then without much effort

tasks/components/Button.tsx
```diff
@@ -52,7 +52,7 @@ const Button = ({
       // primary & {accent, accent-secondary, positive, negative, neutral}
       'text-white bg-indigo-500 hover:bg-indigo-600 focus:ring-indigo-500': variant === 'primary' && color === 'accent', // TODO use hw-primary instead of indigo
       'text-white bg-AAA-500    hover:bg-AAA-600    focus:ring-AAA-500': variant === 'primary' && color === 'accent-secondary', // TODO: what could this be?
-      'text-white bg-XXX-500    hover:bg-XXX-600    focus:ring-XXX-500': variant === 'primary' && color === 'positive', // TODO: green
+      'text-white bg-hw-positive-500    hover:bg-hw-positive-600    focus:ring-hw-positive-500': variant === 'primary' && color === 'positive', // TODO: green
```
JannikWibker, line 55: remove the todo comment as well

tasks/components/Pagination.tsx
```diff
@@ -0,0 +1,64 @@
+import { tw, tx } from '@helpwave/common/twind/index'
+import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
+import { useTranslation } from '@helpwave/common/hooks/useTranslation'
+import type { Languages } from '@helpwave/common/hooks/useLanguage'
+import ChevronLast from '@helpwave/common/icons/ChevronLast'
+import ChevronLeft from '@helpwave/common/icons/ChevronLeft'
+import ChevronFirst from '@helpwave/common/icons/ChevronFirst'
+import ChevronRight from '@helpwave/common/icons/ChevronRight'
+
+type PageIndicatorTranslation = {
```
JannikWibker, line 10: update the name here as well (there are still a few other instances using `pageIndicator`)

tasks/components/OrganizationMemberList.tsx
```diff
@@ -0,0 +1,257 @@
...
+]
+
+export const OrganizationMemberList = ({
+  language,
+  usersPerPage = 5,
+  organization
+}: PropsWithLanguage<OrganizationMemberListTranslation, OrganizationMemberListProps>) => {
+  const translation = useTranslation(language, defaultOrganizationMemberListTranslations)
+  // State copy of the members, so discarding changes is possible
+  const [members, setMembers] = useState(organization.members.map((value) => {
```
JannikWibker, line 102: why are you doing the `.map` here?

DasProffi: To create a copy of the member list, so the status quo can be maintained, if changes are not saved or discarded and to not modify the original objects.

If you have a better way of create copys please let me know

JannikWibker: you don't need to copy it, the original object is never modified

DasProffi: If we were to allow the ability to change roles the object would be modified or does the table itself create a copy, but I will remove the mapping for now

JannikWibker: The object reference you pass into `useState` is never written to, because you always use `setMembers` to update the state.

JannikWibker: The only concern with regards to references and react is "cache invalidation"
You never modify the reference you receive, that just doesn't happen with react and is therefore no concern.

What is however a concern is wrongful invalidation and subsequent re-renders which don't need to happen

tasks/components/OrganizationMemberList.tsx
```diff
@@ -0,0 +1,255 @@
...
+]
+
+export const OrganizationMemberList = ({
+  language,
+  usersPerPage = 5,
+  organization
+}: PropsWithLanguage<OrganizationMemberListTranslation, OrganizationMemberListProps>) => {
+  const translation = useTranslation(language, defaultOrganizationMemberListTranslations)
+  // State copy of the members, so discarding changes is possible
+  const [members, setMembers] = useState([...organization.members])
```
JannikWibker, line 102: you are still copying values here?

just do `useState(organization.members)`

JannikWibker (approved):

PR #120, opened 2023-03-21
extend card component and apply it

### LINKED ISSUE #119: Use Card Component
As mentioned in #112 the `Card` will be used for #110 and #111

DasProffi:
closes #119


app/components/Card.tsx
```diff
@@ -1,19 +1,22 @@
...
   children,
   isSelected = false,
-  onTileClicked = () => undefined
+  onTileClicked = () => undefined,
+  classes = [],
 }: PropsWithChildren<CardProps>) => {
   return (
     <div onClick={onTileClicked}
-         className={tx('cursor-pointer rounded-md py-2 px-4 border border-2 hover:border-hw-primary-700 w-full', { 'border-hw-primary-700': isSelected })}>
+         className={tx('cursor-pointer rounded-md py-2 px-4 border border-2 hover:border-hw-primary-700 w-full', { 'border-hw-primary-700': isSelected }, ...classes)}>
```
JannikWibker, line 19: I'm not always that much of a fan of exposing `className` to the outside

Regardless, I think this should just be a `string`, if the consumer of this component wants to use twind/tailwind classes they should call `tw` or `tx` themself and pass the return value of that along

app/components/Card.tsx
```diff
@@ -1,19 +1,22 @@
 import { tx } from '@helpwave/common/twind/index'
 import type { PropsWithChildren } from 'react'
+import type { Class } from '@twind/core'
 
 export type CardProps = {
   isSelected?: boolean,
-  onTileClicked?: () => void
+  onTileClicked?: () => void,
```
JannikWibker, line 7: I made a mistake in the last (2) review(s): should be `onTileClick` instead of `onTileClicked`

Sorry, was tired 😅

app/components/WardTile.tsx
```diff
@@ -29,10 +31,8 @@ type WardDTO = {
 
 export type WardTileProps = {
   ward: WardDTO,
-  isSelected: boolean,
-  onTileClicked?: () => void,
   onEditClicked?: () => void
-}
+} & CardProps
```
JannikWibker, line 35: When doing `type ABC = { ... } & D` I usually prefer to have the `D` come first, meaning `type ABC = D & { ... }`, or applied to this specific case:

```ts
export type WardTileProps = CardProps & {
  ward: WardDTO,
  onEditClick?: () => void
}
```

app/components/Card.tsx
MaxSchaefer: But why are you not a fan of exposing `className`? What are the downsides?

MaxSchaefer (approved):

app/components/Card.tsx
JannikWibker: Exposing `classNames` often leads to hacky and diverging styling (w.r.t the original design) if used in different places and gets overused quite a lot. If you want some special style you should think really hard about it and maybe add a variant to the original component.

There are of course occasions where exposing this makes a lot of sense but it gets used too much and without much thought too often.

app/components/Card.tsx
```diff
@@ -3,17 +3,19 @@ import type { PropsWithChildren } from 'react'
 
 export type CardProps = {
   isSelected?: boolean,
-  onTileClicked?: () => void
+  onTileClick?: () => void,
+  classes?: string
```
JannikWibker, line 7: I'm still not sold on the idea of passing through the classname here but regardless:
this should be called `className` instead of `classes`

app/components/Card.tsx
JannikWibker: @MaxSchaefer 
(pinging you as this got marked as resolved and I'm not sure if you'd get notified about my answer without a ping)

PR #123, opened 2023-03-25
Fix code style

JannikWibker:
The code style got pretty out of control as of recently and our current linting setup isn't detecting it. Thus this PR adds prettier which is much stricter in enforcing and fixing coding style issues.

### DIFF
--- a/tasks/pages/forgot-password.tsx
+++ b/tasks/pages/forgot-password.tsx
@@ -1,11 +1,7 @@
 import type { NextPage } from 'next'
 
 const ForgotPasswordPage: NextPage = () => {
-  return (
-    <div>
-      not sure how this is going to look or even work, just a placeholder right now
-    </div>
-  )
+  return <div>not sure how this is going to look or even work, just a placeholder right now</div>
 }
 
 export default ForgotPasswordPage
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,14 +3,25 @@ module.exports = {
     'standard',
     'plugin:react/recommended',
     'plugin:@typescript-eslint/recommended',
-    'plugin:react-hooks/recommended'
+    'plugin:react-hooks/recommended',
+    'plugin:prettier/recommended'
   ],
   parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaVersion: '2020'
   },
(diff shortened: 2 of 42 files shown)


tasks/pages/forgot-password.tsx
```diff
@@ -1,11 +1,7 @@
 import type { NextPage } from 'next'
 
 const ForgotPasswordPage: NextPage = () => {
-  return (
-    <div>
-      not sure how this is going to look or even work, just a placeholder right now
-    </div>
-  )
+  return <div>not sure how this is going to look or even work, just a placeholder right now</div>
```
MaxSchaefer, line 4: We should at least stick with the leading and trailing brackets.

MaxSchaefer (changes requested):
My two cents; We should keep:
- Commas overall (especially for `type`)
- Trailing commas, to prevent unnecessary git additions

JannikWibker:
@MaxSchaefer I kind of agree with both your comments, but there is one problem with this:
- prettier is pretty opinionated and doesn't really allow for rules to be enforced in a non-strict way; meaning if you "allow" trailing commas you need them everywhere, there is no in between and taking that into account i'm more on the "against"-side when it comes to trailing commas
- prettier is light years ahead of other code formatting and linting tools, I used to dislike prettier because of the highly opinionated and non-configurable nature but have grown to rely on it really much as no other tool detects and fixes this many issues; thus not using prettier is in my opinion not really an option

MaxSchaefer:
> @MaxSchaefer I kind of agree with both your comments, but there is one problem with this:
> 
> * prettier is pretty opinionated and doesn't really allow for rules to be enforced in a non-strict way; meaning if you "allow" trailing commas you need them everywhere, there is no in between and taking that into account i'm more on the "against"-side when it comes to trailing commas
> * prettier is light years ahead of other code formatting and linting tools, I used to dislike prettier because of the highly opinionated and non-configurable nature but have grown to rely on it really much as no other tool detects and fixes this many issues; thus not using prettier is in my opinion not really an option

We should not adopt code styles because the tools cannot be configured to our liking. I've skimmed over the changes, with a stricter configuration, ESLint should be able to handle this as well.

use-to:
Houston, we have a problem!

PR #125, opened 2023-03-28
add ProgressIndicator

### LINKED ISSUE #124: ProgressIndicator
As mentioned in #96 this is a seperate issue for the ProgressIndicator

DasProffi:
closes #124
Unbenannt

### DIFF
--- a/tasks/components/ProgressIndicator.tsx
+++ b/tasks/components/ProgressIndicator.tsx
@@ -1,10 +1,19 @@
 import { tw } from '@helpwave/common/twind/index'
 
 export type ProgressIndicatorProps = {
+  /*
+    The amount of progress that has been made
+    Value form 0 to 1
+   */
   progress: number,
   strokeWidth?: number,
   size?: keyof typeof sizeMapping,
   direction?: 'clockwise' | 'counterclockwise',
+  /*
+    Rotation of the starting point of the indicator
+    default start at 3 o'clock
+    Given in degree
+   */
   rotation?: number
 }
 
--- a/tasks/components/ProgressIndicator.tsx
+++ b/tasks/components/ProgressIndicator.tsx
@@ -1,0 +1,32 @@
+import { tw } from '@helpwave/common/twind/index'
+
+export type ProgressIndicatorProps = {
+  progress: number,
+  strokeWidth?: number,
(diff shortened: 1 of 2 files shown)


MaxSchaefer:
> @MaxSchaefer do you have any idea why our ci fails?

`#17 0.854  ERR_PNPM_FROZEN_LOCKFILE_WITH_OUTDATED_LOCKFILE  Cannot perform a frozen installation because the lockfile needs updates`

The `pnpm-lock.yaml` is out-of-sync.

https://github.com/pnpm/pnpm/issues/2807#issuecomment-682471525

DasProffi:
> [pnpm/pnpm#2807 (comment)](https://github.com/pnpm/pnpm/issues/2807#issuecomment-682471525)
Doesn't work, `pnpm install` doesn't change the lock file, I tried this even with completly deleting the local repository.

Could someone else verify that the lock file isn't updated and if it does, push the commit? @MaxSchaefer

MaxSchaefer:
Of course, we should also use a current PNPM version in the container. Why it has worked for us in CI so far, good question. Since PNPM 8 came out today, and we were on version 6, I assume that this has contributed to the fail of the CI.

tasks/components/ProgressIndicator.tsx
```diff
@@ -0,0 +1,32 @@
+import { tw } from '@helpwave/common/twind/index'
+
+export type ProgressIndicatorProps = {
+  progress: number,
+  strokeWidth?: number,
+  size?: 'small' | 'medium' | 'big',
```
MaxSchaefer, line 6:
```suggestion
  size?: keyof typeof sizeMapping,
```

tasks/components/ProgressIndicator.tsx
```diff
@@ -0,0 +1,32 @@
+import { tw } from '@helpwave/common/twind/index'
+
+export type ProgressIndicatorProps = {
+  progress: number,
```
MaxSchaefer, line 4: Please add some comments of the expected values. Like 0 to 1 for progress.

MaxSchaefer (changes requested):

MaxSchaefer:
Nice job, I like the SVG approach. 👍

MaxSchaefer (approved):

PR #139, opened 2023-04-01
add kanban component

### LINKED ISSUE #96: Kanban component for tasks
Tasks Section

This component should be an aggregate of many other components like `status pills`, `single task tile`, `kanban header` and also the `kanban columns`. It should be possible to pass an array of tasks to the component.

DasProffi:
closes #96
Capture

### DIFF
--- a/tasks/components/KanbanHeader.tsx
+++ b/tasks/components/KanbanHeader.tsx
@@ -1,0 +1,58 @@
+import { tw } from '@helpwave/common/twind/index'
+import type { Languages } from '@helpwave/common/hooks/useLanguage'
+import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
+import { useTranslation } from '@helpwave/common/hooks/useTranslation'
+import TriangleDown from '../icons/TriangleDown'
+import { Input } from './Input'
+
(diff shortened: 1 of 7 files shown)


tasks/components/KanbanHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+  sortingStatus?: string,
+  sortingLabel?: string,
+  searchValue: string,
+  onSearchChange: (search: string) => void
+}
+
+export const KanbanHeader = ({
+  language,
+  sortingStatus,
+  sortingLabel,
```
MaxSchaefer, line 37: Both props are not in use

MaxSchaefer (changes requested):
Works great! ✌️
We can improve the UX in the future by replacing some colors and animations.

tasks/components/KanbanHeader.tsx
DasProffi: because we haven't establish a sorting for the status or label? But the ui indicated the possibility

DasProffi: I will just omit them for now in the function props

MaxSchaefer (approved):

PR #194, opened 2023-04-13
add ToggleableInput and use it in patient form

### LINKED ISSUE #189: Edit patient name on details view
On the details view, the patient name should be inline editable. Checkout #174 for an edit icon.

room headings variant 3

DasProffi:
closes #189

### DIFF
--- a/lib/icons/Edit.tsx
+++ b/lib/icons/Edit.tsx
@@ -1,0 +1,22 @@
+import React, { forwardRef } from 'react'
+import type { SVGProps } from 'react'
+
+export default forwardRef<SVGSVGElement, SVGProps<SVGSVGElement>>(function Edit(props, ref) {
+  return (
+    <svg
+      width="24"
+      height="24"
+      viewBox="0 0 24 24"
+      fill="none"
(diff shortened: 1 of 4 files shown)


MaxSchaefer (changes requested):
Can we make sure that there is now layout shift by reducing the padding around the input?

https://user-images.githubusercontent.com/40192999/231895978-e54d71a6-77a9-45bc-8153-b24ab56fcfe3.mov

MaxSchaefer (changes requested):
Now the text itself gets shifted and is bold during editing. Can we make this a truly __inline edit__? The purple input border should be around the text without the text-shift. The user should be able to confirm the input field also via the enter key.

https://user-images.githubusercontent.com/40192999/232250277-746e1fa0-c8ad-4af0-8a42-f49389c0f7d5.mov

MaxSchaefer:
As discussed, we will replace the border with an underline during the edit mode.

MaxSchaefer (approved):
LGTM! I've added an `underline-offset-4` to gain more space between the text and the underline.

Issue #183, opened 2023-04-12
Remove needless pages and components


use-to:
There are some needless components:
- [x] Sign-In and Sign-Up (as this will be handled by ory.sh)
- [x] ContactPage `/contact`
- [x] Forget Password
- [x] SettingsPage #188 

and so on.
Please also create a branch called `glacier/old-pages` providing the current version.

PR #204, opened 2023-04-15
fix: use consistent title

### LINKED ISSUE #202: Add proper page title
`Create Next App` ...

MaxSchaefer:
Use `helpwave tasks` or the suffix `~ helpwave tasks`

Closes #202

### DIFF
--- a/tasks/pages/profile.tsx
+++ b/tasks/pages/profile.tsx
@@ -10,7 +10,7 @@ import { Input, noop } from '../components/user_input/Input'
 import { Avatar } from '../components/Avatar'
 import { UserCard } from '../components/cards/UserCard'
 import { PageWithHeader } from '../components/layout/PageWithHeader'
-import titleWrapper from '../utils/titleWrapper';
+import titleWrapper from '../utils/titleWrapper'
 
 const Section = ({ title, children }: PropsWithChildren<{ title: string }>) => (
   <div className={tw('flex')}>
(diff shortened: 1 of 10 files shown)


florian-sabonchi (commented):
I think we should vote again whether to take a pipe or a tilde

tasks/pages/profile.tsx
```diff
@@ -54,7 +54,7 @@ const ProfilePage: NextPage = () => {
   return (
     <PageWithHeader>
       <Head>
-        <title>Settings</title>
```
DasProffi, line 57: profile should be translated

tasks/pages/settings.tsx
```diff
@@ -28,7 +28,7 @@ const SettingsPage: NextPage = () => {
   return (
     <PageWithHeader>
       <Head>
-        <title>Settings</title>
```
DasProffi, line 31: settings needs to be translated

DasProffi (changes requested):

DasProffi:
Rather than hardcoding the suffix on every page we should use a function to wrap the name of the page
```
const titleWrapper = (title: string) => title + '~ helpwave tasks'
```
This should allow us to easily modify it

tasks/pages/profile.tsx
use-to: profile page is legacy code #183

MaxSchaefer:
@DasProffi @florian-sabonchi
` ~ ` yes or no? The alternatives are ` | ` or ` - `.

DasProffi (approved):

PR #206, opened 2023-04-15
add reusable modal

### LINKED ISSUE #199: Add Modal View
Add a package or component that allows showing it's children as a modal on the top level

DasProffi:
closes #199

### DIFF
--- a/tasks/components/modals/ConfirmDialog.tsx
+++ b/tasks/components/modals/ConfirmDialog.tsx
@@ -63,7 +63,7 @@ export const ConfirmDialog = ({
             {translation.decline}
           </Button>
         )}
-        <Button autoFocus color="positive" onClick={() => onConfirm()}>
+        <Button autoFocus color="positive" onClick={onConfirm}>
           {translation.confirm}
         </Button>
       </div>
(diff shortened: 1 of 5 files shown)


use-to (changes requested):
1. modal overlay - click event (dismiss function) should be optional
2. modal overlay - background should be darker
3. fix `onClick`-event on actual modal container (currently triggers `onClose`)

tasks/components/modals/Modal.tsx
JannikWibker, line 1: could also be moved to `lib/`

tasks/components/modals/ConfirmDialog.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <Button color="accent" onClick={event => {
+          event.stopPropagation()
+          onClose('closed')
+        }}>
+          {translation.cancel}
+        </Button>
+        {isShowingDecline && (
+          <Button color="negative" onClick={event => {
+            event.stopPropagation()
+            onClose('decline')
```
JannikWibker, line 60: I'd vouch for having individual callbacks for each action (`'onDecline'`, `onAccept` or `onConfirm`, `onClose`)

But I'm open to discussing this; if there is something specific this allows I'm all for it

tasks/components/modals/Modal.tsx
use-to: I would say there are many components we could move to the `lib/` directory.
Maybe we should do it later on?

DasProffi: I will make an issue for that #212

use-to (approved):

PR #235, opened 2023-04-20
improve search of tasks

### LINKED ISSUE #216: Improve PatientDetails Search
- [ ] add hint text to search
- [ ] improve search (currently the first characters have to match in order to score a result)

DasProffi:
closes #216


MaxSchaefer (approved):

use-to (approved):
I would prefer `includes` over `indexOf` next time although `includes` is a very very bit slower because it checks for regex input...

use-to:
We'll improve this later on anyway because we need more precise and fault-tolerant search services.

PR #242, opened 2023-04-20
remove unnecessary console.log

### LINKED ISSUE #237: Remove obsolete console.log
there are several console.log withing the code we should remove them

DasProffi:
closes #237


DasProffi:
What do we want to do with other console.info? Should we also remove them and any code that isn't production code

use-to:
> 

I'm not entirely sure how relevant the current logging messages are...
But in general it is good practice to log in anyway relevant behavior of our app.
Big applications like discord are using `console.log` and `console.info` statements, too.

MaxSchaefer (approved):

PR #243, opened 2023-04-20
Apply Confirm Dialog

### LINKED ISSUE #215: Apply ConfirmDialog
Use the Confirm dialog as the default dialog instead of a confirm.

depends on #199

use-to:
Closes #215

### DIFF
--- a/tasks/components/RoomList.tsx
+++ b/tasks/components/RoomList.tsx
@@ -1,3 +1,6 @@
+import type {
+  Cell, CoreCell
+} from '@tanstack/react-table'
 import {
   useReactTable,
   createColumnHelper,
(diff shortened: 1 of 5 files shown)


tasks/components/RoomList.tsx
```diff
@@ -32,7 +40,10 @@ const defaultRoomListTranslations: Record<Languages, RoomListTranslation> = {
     selectAll: 'Select All',
     rooms: 'Rooms',
     addRoom: 'Add Room',
-    bedCount: 'Number of Beds'
+    bedCount: 'Number of Beds',
+    dangerZoneText: 'Deleting room(s) is a permanent action and cannot be undone. Be careful!',
+    deleteConfirmText: 'Do you really want to delete the selected room(s)?',
+    deleteRoom: 'Delete Room(s)',
```
MaxSchaefer, line 46: We should not use pluralize via an `(s)`. Most of the time, we already have the context of how many entities this action will affect. In that case, there is also already the singularized `room`.

MaxSchaefer (changes requested):

MaxSchaefer (approved):

PR #250, opened 2023-04-20
add SignInButton

### LINKED ISSUE #234: `Sign in with helpwave`-Button
The [current implementation](https://github.com/helpwave/web/blob/cb3fe15cc56fd424714f7de0acdd068f01a51434/tasks/components/SignInButton.tsx) should be modified to match the mocks.

Frame 25

DasProffi:
closes #234


lib/components/SignInButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    signIn: 'Sign in with helpwave'
+  },
+  de: {
+    signIn: 'Anmelden mit helpwave'
+  }
+}
+
+export type SignInButtonProps = {
+  color?: 'dark' | 'light',
+  onClick: () => void
```
MaxSchaefer, line 22: In the future, we will provide a default behavior.

MaxSchaefer: Please provide a default value like `no-op`.

MaxSchaefer (changes requested):

tasks/components/SignInButton.tsx
```diff
@@ -1,32 +0,0 @@
...
-}
-
-const SignInButton = ({
-  language,
-}: PropsWithLanguage<SignInButtonTranslation>) => {
-  const translation = useTranslation(language, defaultSignInButtonTranslation)
-
-  return (
-    <Button variant="secondary" onClick={() => {
-      getAuthorizationUrl().then((url) => window.location.assign(url))
```
MaxSchaefer, line 27: Can you add this to the `onClick` handler on `pages/auth/index.tsx`? When this pr introduces changes to the current SignInButton, the behavior should stay the same.

MaxSchaefer (changes requested):

Issue #245, opened 2023-04-20
Replace green confirm button for deletion with red


florian-sabonchi:
![image](https://user-images.githubusercontent.com/54689374/233423435-cad3fc9a-c232-43f7-9b3d-36ffc10c3d5e.png)

shivadharmi:
Hey, @florian-sabonchi I would like to work on this issue. Can I go for it?

florian-sabonchi:
Yes of course

florian-sabonchi:
@shivadharmi you can work on it a list of our code convetions can be found here https://github.com/helpwave/helpwave/blob/main/development/CONVENTIONS.md

shivadharmi:
Thank you @florian-sabonchi.

shivadharmi:
Hi @florian-sabonchi. I have gone through the contribution documentation. I did the project setup in my local. I might need credentials to log in. Can you help me with this?

florian-sabonchi:
> Hi @florian-sabonchi. I have gone through the contribution documentation. I did the project setup in my local. I might need credentials to log in. Can you help me with this?

Hi, thanks for your message  there is no login you can skip this step by entering empty username and password

shivadharmi:
Hey @florian-sabonchi . As I can see from the code we are using a confirmation dialogue component.
There can be three types of actions:
- `Sucess`
- `Warning` 
- `Danger`
Shouldn't we consider three of them and design the confirmation dialogue in such a way that we can use it in all three scenarios? Where we will show red for negative actions, green for positive, and yellow for neutral.

florian-sabonchi:
Yes, i agree sounds food to me

shivadharmi:
okay @florian-sabonchi. I will work on it. Thanks!

use-to:
> Hey @florian-sabonchi . As I can see from the code we are using a confirmation dialogue component. There can be three types of actions:
> 
> * `Sucess`
> * `Warning`
> * `Danger`
>   Shouldn't we consider three of them and design the confirmation dialogue in such a way that we can use it in all three scenarios? Where we will show the button in red color for negative actions, green color for positive, and yellow color for neutral.

Hey @shivadharmi, welcome!
Feel free taking a look at our design definitions (we got some colors there too):
https://www.figma.com/file/vCeohZgfJzkN4VQreBvAM0/helpwave-team-library?node-id=0%3A1&t=KXHkZ1jWg1mm3N0K-1
We received your mail and currently processing it!
[...]

PR #255, opened 2023-04-21
feat: replace green confirm button with red color

shivadharmi:
This pull request replaces the green confirm button with red color

Closes https://github.com/helpwave/web/issues/245

Testing instructions:

- Go to the organizations' page and click the edit icon of any organization from the list to see organization info.
- Under the `Danger Zone` section click on the `Delete Organization` button you will see a confirmation dialogue and it should contain a confirmation button in red color instead of green.

Additional notes:

- Refactored the confirmation dialogue to support different types of confirmations. The default confirmation dialogue is `Positive Action`.

### DIFF
--- a/tasks/components/modals/ConfirmDialog.tsx
+++ b/tasks/components/modals/ConfirmDialog.tsx
@@ -12,11 +12,7 @@ type ConfirmDialogTranslation = {
   decline: string
 }
 
-export enum ConfirmDialogType{
-  Positive = 'Positive',
-  Neutral = 'Neutral',
-  Negative = 'Negative'
-}
+export type ConfirmDialogType = 'positive' | 'negative' | 'neutral'
 
 const defaultConfirmDialogTranslation = {
   en: {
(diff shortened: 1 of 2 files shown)


tasks/components/modals/ConfirmDialog.tsx
```diff
@@ -12,6 +12,12 @@ type ConfirmDialogTranslation = {
   decline: string
 }
 
+export enum ConfirmDialogType{
+  Positive = 'Positive',
+  Neutral = 'Neutral',
+  Negative = 'Negative'
+}
```
DasProffi, line 19: Rather use a string union than an actual enum
```suggestion
export type ConfirmDialogType = 'positive' | 'negative' | 'neutral'
```

DasProffi (changes requested):

shivadharmi:
Hey, @DasProffi I would like to delete this PR. As my commit was done from my office email. Sorry for the trouble. Is that okay?

PR #273, opened 2023-04-22
add DiscardChangesDialog

### LINKED ISSUE #264: Add discard changes dialog
```[tasklist]
### Tasks
- [ ] Add a discard changes dialog
- [ ] apply this dialog to the organization, ward and patient page
```

The dialog should afford the options discard, change and close, keep editing

DasProffi:
closes #264

### DIFF
--- a/tasks/components/layout/OrganizationDetails.tsx
+++ b/tasks/components/layout/OrganizationDetails.tsx
@@ -59,7 +59,7 @@ export const OrganizationDetail = ({
   setOrganization
 }: PropsWithLanguage<OrganizationDetailTranslation, OrganizationDetailProps>) => {
   const translation = useTranslation(language, defaultOrganizationDetailTranslations)
-  const isCreatingNewOrganization = organizationForm.organization.id !== undefined
+  const isCreatingNewOrganization = organizationForm.organization.id === ''
   const [isShowingConfirmDialog, setIsShowingConfirmDialog] = useState(false)
 
   return (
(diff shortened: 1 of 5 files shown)


DasProffi:
Currently, only for the organizations, do we want to apply this to the other screens as well in the same manner or do we want to change the handling?

use-to:
> Currently, only for the organizations, do we want to apply this to the other screens as well in the same manner or do we want to change the handling?

Good question, we should take a look at this.
Lets move this for now and create a new issue for that application.

florian-sabonchi:
Bildschirmfoto vom 2023-04-24 21-27-49
I would  suggest from an ui/ux perspektive to remove the cancel button

tasks/components/layout/OrganizationDetails.tsx
```diff
@@ -101,23 +72,35 @@ export const OrganizationDetail = ({
...
+          members={organizationForm.organization.members}
+          onChange={(members) => setOrganization((prevState) => {
+            return {
+              hasChanges: true,
+              isValid: prevState.isValid,
+              organization: {
+                ...prevState.organization,
+                members
+              }
+            }
```
use-to, line 102: Btw there is a trick for this. You can do `() => ({key: "value"})`

use-to (changes requested):

tasks/components/layout/OrganizationDetails.tsx
DasProffi: in general yes, but not in this case because the previousStates organization is required, or how would you like to resolve this?

shivadharmi (approved):

tasks/components/layout/OrganizationDetails.tsx
MaxSchaefer:
```js
onChange={(members) => setOrganization((prevState) => ({ isValid: prevState.isValid, test: 'test' }))}
```

use-to (approved):

PR #301, opened 2023-04-25
redesign room and member list

### LINKED ISSUE #269: Make Member/Room lists header prettier
- those tables are ugly af
- it's not possible to differentiate between header title and button (with action) like remove
- For member list: at the moment we currently unable to change any roles on a member (just disabled this selection indicator thing - what so ever...)

DasProffi:
closes #269

### DIFF
--- a/tasks/components/OrganizationMemberList.tsx
+++ b/tasks/components/OrganizationMemberList.tsx
@@ -6,7 +6,6 @@ import {
   getPaginationRowModel
 } from '@tanstack/react-table'
 import { tw } from '@helpwave/common/twind'
-import { ChevronDown } from 'lucide-react'
 import type { Languages } from '@helpwave/common/hooks/useLanguage'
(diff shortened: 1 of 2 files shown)


MaxSchaefer (changes requested):
I would suggest for all lists an horizontal line between the header and table content section. 

Also please address the miss-alignment of the following header checkbox:
Bildschirm­foto 2023-04-25 um 9 05 34 PM

MaxSchaefer (changes requested):
Looks good, but the horizontal line is not continues. I guess we should move the line before the body.
Bildschirm­foto 2023-04-25 um 10 48 16 PM

DasProffi:
> Looks good, but the horizontal line is not continues. I guess we should move the line before the body. Bildschirm­foto 2023-04-25 um 10 48 16 PM

For me it is continuous and moving it to the body makes it only go the length of the first cell

MaxSchaefer (approved):
The previous issue was fixed! Thanks!

PR #313, opened 2023-04-27
destructure props and rename translations

### LINKED ISSUE #291: Destructure component props when used
The following component use not destructured props:
- `PillLabel`

DasProffi:
closes #291


MaxSchaefer (approved):
LGTM! But I like the generic `FunctionComponent` more 😄

PR #315, opened 2023-04-28
add base component for uniform textStyle

### LINKED ISSUE #292: Uniform Text style
There are currently around 10 different text styles maybe more

We should create a component or special helpwave tailwind styles for that and then apply them

DasProffi:
closes #292


MaxSchaefer (dismissed):
I like the new `Span` component 🤘

PR #338, opened 2023-05-02
Issue/271 persist lang setting

### LINKED ISSUE #271: Persist Language
> We should persist the language selection.
Maybe we could add this to the current jwt token.

- [ ] Use `window.localStorage` to persits the current selected language

Paul-B98:
closes #271


MaxSchaefer (changes requested):
In React-Land we should move this inside a Hook. What speaks against a simplified version of an [`useLocalStorage()`](https://usehooks-ts.com/react-hook/use-local-storage) or [`use-persisted-state`](https://github.com/donavon/use-persisted-state)? We can wrap `LocalStroageService` with such a hook.

use-to:
@Paul-B98 I think @MaxSchaefer wants you to throw the `StorageSerice` away.

lib/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,41 @@
...
+declare global {
+  interface WindowEventMap {
+    'local-storage': CustomEvent
+  }
+}
+
+type SetValue<T> = Dispatch<SetStateAction<T>>
+type FunctionType = <T>(key: string, initValue: T) => [T, SetValue<T>]
+
+const useLocalStorage: FunctionType = <T, >(key: string, initValue: T) => {
```
MaxSchaefer, line 14:
```suggestion
const useLocalStorage = <T, >(key: string, initValue: T): [T, SetValue<T>] => {
```

lib/hooks/useLocalStorage.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const get = useCallback((): T => {
+    if (typeof window === "undefined") {
```
MaxSchaefer, line 16:
```suggestion
    if (typeof window === 'undefined') { 
```

MaxSchaefer (changes requested):

MaxSchaefer:
> @Paul-B98 I think @MaxSchaefer wants you to throw the `StorageSerice` away.

No, not necessarily. It's just an unusually OOP approach in our Frontend, the `hook`-Wrapper translates this very well into React-Land.

MaxSchaefer (approved):
@Paul-B98 looks good to me! Good job
I've just moved the call to `setStoredLanguage` directly into the `useLanguage`-Hook. Seams like a good idea that every call to `setLanguage` should persist the language as well.

MaxSchaefer:
@Paul-B98 I've invited you to the web-Team. You can consider a squash merge.

PR #342, opened 2023-05-04
fix: swap create/update button with danger zone

### LINKED ISSUE #341: Organization Detail View: Change Position of Danger Zone
We should change the position of the danger zone to be under the update button in the organization detail view.

aegis301:
closes #341


use-to (changes requested):
We need to swap it on the WardDetails too.

use-to (dismissed):

Paul-B98 (approved):

MaxSchaefer (approved):

PR #347, opened 2023-05-04
fix: mobile interceptor through media-query instead of MobileView and…

MaxSchaefer:
… BrowserView


MaxSchaefer:
@florian-sabonchi and I figured out that the import in `AvatarGroup` was referencing `@twind/core` instead of our `@helpwave/common/twind` wrapper. This import was the cause of today's issue on main and did not occur during dev.
To prevent such issues, we should prevent such imports and implement something like smoke-tests in our CI.

florian-sabonchi (approved):

Issue #352, opened 2023-05-05
Better Form management


DasProffi:
Currently our form validation and updating is very similar in many components, but there are no components for that

```[tasklist]
### Tasks
- [ ] A FormField with a label, input field, errormessage beneath, state whether it has beed changed
- [ ] A component for state handling, making valdiation, saving state infomation on the data and hasChanged state
```

MaxSchaefer:
@DasProffi how can I help so that the issue becomes ready?

DasProffi:
We should provide a clear solution before making this ready, we have talked about using [react-hook-form](https://react-hook-form.com/), but that isn't tested yet to check whether it is a viable solution. So we either should probably wait for that to open this issue or make it explicit, that this is just an issue to test the viability.

MaxSchaefer:
https://github.com/helpwave/web/issues/370

PR #356, opened 2023-05-05
add auto save to PatientDetails

### LINKED ISSUE #332: Auto-Save
The user should not be required to manually save tasks he or she created for a patient. Right now, if the user does not click "save changes", tasks and other status updates for the patient simply get deleted if the user does not hit "save changes" button. Maybe we should get rid of this button alltogether?

DasProffi:
closes #332

### DIFF
--- a/tasks/components/layout/PatientDetails.tsx
+++ b/tasks/components/layout/PatientDetails.tsx
@@ -2,7 +2,7 @@ import { tw } from '@helpwave/common/twind'
 import type { Languages } from '@helpwave/common/hooks/useLanguage'
 import type { PropsWithLanguage } from '@helpwave/common/hooks/useTranslation'
 import { useTranslation } from '@helpwave/common/hooks/useTranslation'
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import { ColumnTitle } from '../ColumnTitle'
 import { Button } from '@helpwave/common/components/Button'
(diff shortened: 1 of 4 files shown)


tasks/components/layout/PatientDetails.tsx
```diff
@@ -64,14 +73,48 @@ export const PatientDetail = ({
...
+    setUpdateTimer(setInterval(() => {
+      onUpdate(newPatient)
+      // Show Saved Notification for fade animation duration
+      clearInterval(notificationTimer)
+      setNotificationTimer(undefined)
+      setNotificationTimer(setInterval(() => {
+        clearInterval(notificationTimer)
+        setNotificationTimer(undefined)
+      }, 3000))
+    }, 3000))
```
MaxSchaefer, line 104: We need to find a way to generalize this. The debounce and notification logic can be separated. The debounce can be moved to a separate hook, like [useDebounce()](https://usehooks-ts.com/react-hook/use-debounce). With that the notification logic can be handled inside our `useDebounce()` callback.

MaxSchaefer (changes requested):
The potential hook should also have the ability to save directly. For example, when the user leaves the input field, we can save directly.

MaxSchaefer (approved):
Looks good. Feel free to add the ability of save-on-input-defocus in this pr. I guess we will optimize around this with #352 as well.

Issue #288, opened 2023-04-24
Set config defaults for local development


MaxSchaefer:
> Hello @MaxSchaefer. When running the project in my local and visiting the localhost url it is redirecting me to [https://staging-tasks.helpwave.de/auth/callback?fake_token=eyJzdWIiOiIxODE1OTcxMy01ZDRlLT[…]NDc5ODQiLCJlbWFpbCI6InRlc3RpbmUudGVzdEBoZWxwd2F2ZS5kZSJ9](https://staging-tasks.helpwave.de/auth/callback?fake_token=eyJzdWIiOiIxODE1OTcxMy01ZDRlLTRhZDUtOTRhZC1mYmI2YmIxNDc5ODQiLCJlbWFpbCI6InRlc3RpbmUudGVzdEBoZWxwd2F2ZS5kZSJ9)
> Am I missing anything?

You can set the environment variable (`/tasks/.env`) `NEXT_PUBLIC_OAUTH_REDIRECT_URI` to `http://localhost:3000/auth/callback`.

_Originally posted by @MaxSchaefer in https://github.com/helpwave/web/issues/203#issuecomment-1519180447_

MaxSchaefer:
@JaninaWibker wanna take a look at this? Can Next.js 14 help us with runtime envs?

PR #361, opened 2023-05-06
chore: set auth envs for staging

MaxSchaefer:
This pull request sets the envs for auth during the build process. This resolves the first to-do of #288


Dockerfile.tasks
```diff
@@ -2,6 +2,9 @@ FROM node:18.16.0 as build
 ENV NEXT_TELEMETRY_DISABLED 1
 ARG WS="@helpwave/tasks"
 
+ENV NEXT_PUBLIC_FAKE_TOKEN_ENABLE=false
+ENV NEXT_PUBLIC_OAUTH_REDIRECT_URI=https://staging-tasks.helpwave.de/auth/callback
```
use-to, line 6: We should pass it using a `ARG`  variable.

tasks/hooks/useAuth.ts
```diff
@@ -70,18 +70,20 @@ export const useAuth = () => {
...
-      setUser(user)
-      return
-    }
+      if (idTokenValid) {
+        const user = tokenToUser(idToken)
+        if (!user) throw new Error('Cannot parse idToken to user')
+        setUser(user)
+        return
+      }
+    } catch (err) {}
```
use-to, line 86:
```suggestion
    } catch (_) {}
```

Should we handle this? Do we have to ensure that some storages are cleared properly?

tasks/utils/config.ts
```diff
@@ -32,7 +32,7 @@ const configSchema = z.object({
   NEXT_PUBLIC_PLAYSTORE_LINK: z.string().url().default('https://play.google.com/store/apps'),
   NEXT_PUBLIC_APPSTORE_LINK: z.string().url().default('https://www.apple.com/de/app-store/'),
   NEXT_PUBLIC_OAUTH_ISSUER_URL: z.string().url().default('https://auth.helpwave.de'),
-  NEXT_PUBLIC_OAUTH_REDIRECT_URI: z.string().url().default('https://staging-tasks.helpwave.de/auth/callback'),
+  NEXT_PUBLIC_OAUTH_REDIRECT_URI: z.string().url().default('http://localhost:3000/auth/callback'),
```
use-to, line 35: This is okay but maybe we should provide an addtional `tasks/.env.local` containing this setting.

use-to (changes requested):

tasks/utils/config.ts
MaxSchaefer: Okay, the defaults should be "production" ready. The `.env` defines the default for the local development setup. The docker build gets not influenced by the `.env` because of the `.dockerignore`. Right now, the only way to influence the build envs are `.env.local`.

tasks/hooks/useAuth.ts
MaxSchaefer: Totally! We should not ignore this error 😄

use-to (approved):
Doable 👍🏻

PR #372, opened 2023-05-07
split template mutations

### LINKED ISSUE #364: Personal and Ward Task Templates uses same `queryKey`
As the title already implies the personal and ward task template screen uses the same `queryKey` in React Query.
That leads to the same items displayed in that list.

DasProffi:
closes #364


tasks/mutations/personal_task_template_mutations.ts
```diff
@@ -10,36 +10,54 @@ export type TaskTemplateDTO = {
   isPublicVisible: boolean
 }
 
-const queryKey = 'taskTemplates'
+type QueryKey = 'personalTaskTemplates'| 'wardTaskTemplates'
```
use-to, line 13: Constant definitions would be useful...
`const personalTaskTemplates = "personalTaskTemplates"` ?

use-to (changes requested):

tasks/mutations/personal_task_template_mutations.ts
DasProffi: but then you can't really use them as a parameter or do you mean something like: `const queryKey = ['personalTaskTemplate', 'wardTaskTemplate]`

use-to (approved):

PR #386, opened 2023-05-10
outline overdue tasks

### LINKED ISSUE #383: Red outline for overdue tasks
To indicate overdue tasks, the first approach in our UI, should be to strongly outline these tasks with a red border.

DasProffi:
closes #383

### DIFF
--- a/lib/components/Card.tsx
+++ b/lib/components/Card.tsx
@@ -1,10 +1,11 @@
 import { tx } from '../twind'
 import type { PropsWithChildren } from 'react'
+import type { Class } from '@twind/core'
 
 export type CardProps = {
   isSelected?: boolean,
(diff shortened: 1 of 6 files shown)


Paul-B98:
hey, maybe we should sync our pull request? in #387 are some change that introduce a date picker

DasProffi:
> hey, maybe we should sync our pull request? in #387 are some change that introduce a date picker

Makes sense, I will merge your PR into this one, so you can close yours

use-to (changes requested):
We definitely need time as well. So can we pick a datetime instead of date?

DasProffi:
In the long run, we should probably consider another component just for date/datetime input or search packages that do so

use-to (approved):

PR #406, opened 2023-05-13
Issue/392 input disclaimer

### LINKED ISSUE #392: Input disclaimer
All of our components where the user can input arbitrary text should provide something like an "input hint". The input hint text should only visible when the user is about to use the input element. In the future, we will use this input element to indicate sensitive fields where the user should not enter patient data.

```[tasklist]
### Implementing input hint
- [ ] [Input.tsx](https://github.com/helpwave/web/blob/main/lib/components/user_input/Input.tsx)
- [ ] [Textare.tsx](https://github.com/helpwave/web/blob/main/lib/components/user_input/Input.tsx)
- [ ] Adopting "input hint" for tasks
```

Bildschirm­foto 2023-05-11 um 10 48 42 PM

florian-sabonchi:
closes #392

image
image

### DIFF
--- a/lib/components/user_input/Textarea.tsx
+++ b/lib/components/user_input/Textarea.tsx
@@ -47,7 +47,7 @@ export const Textarea = ({ headline, id, resizable = false, onChange = noop, dis
         >
       </textarea>
       </div>
-      {hasFocus && (
+      {(hasFocus && disclaimer) && (
         <label className={tw('text-hw-negative-500')}>
           {disclaimer}
         </label>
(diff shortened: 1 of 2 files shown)


use-to (approved):
It's implemented as described in the issue. Good job!
We maybe have do deal with text overflows resulting into layout problems.

use-to:
I would doubt that this is an actual [chore](https://github.com/helpwave/web/pull/406/commits/6a7fe798154b3bfecb2c0c4cf1242e0f97714bd1) for us.

PR #432, opened 2023-05-19
fix: empty inputs

### LINKED ISSUE #426: empty names
We should check for empty names to avoid erroing occuring through this
image

florian-sabonchi:
closes #426

### DIFF
--- a/tasks/components/OrganizationForm.tsx
+++ b/tasks/components/OrganizationForm.tsx
@@ -97,7 +97,7 @@ export const OrganizationForm = ({
   const inputClasses = tw('mt-1 block rounded-md w-full border-gray-300 shadow-sm focus:outline-none focus:border-hw-primary-500 focus:ring-hw-primary-500')
 
   function validateShortName(organization: OrganizationGeneralInfoDTO) {
-    if (organization.shortName === '') {
+    if (organization.shortName.trim() === '') {
       return translation.required
     } else if (organization.shortName.length < minShortNameLength) {
       return translation.tooShort(minShortNameLength)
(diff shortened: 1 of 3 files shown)


MaxSchaefer (changes requested):
We should move the trim to the input instead.

DasProffi (approved):

MaxSchaefer (commented):
What's the purpose behind the extra `set[InputAttributeName]` `useState` hook?

florian-sabonchi:
> What's the purpose behind the extra `set[InputAttributeName]` `useState` hook?

Otherwise the user cannot enter spaces, because in onChangeEvengt laws value is passed as value to the input. Thus it would no longer be possible to enter spaces this would be a compromise

MaxSchaefer:
> > What's the purpose behind the extra `set[InputAttributeName]` `useState` hook?
> 
> Otherwise the user cannot enter spaces, because the Value set in the onChangeEvent is is passed as value to the custom input.

Okay, then you can move the `.trim()` logic into the `validate...()` functions. These are called by `triggerOnChange()`.

tasks/components/OrganizationForm.tsx
```diff
@@ -97,7 +97,7 @@ export const OrganizationForm = ({
   const inputClasses = tw('mt-1 block rounded-md w-full border-gray-300 shadow-sm focus:outline-none focus:border-hw-primary-500 focus:ring-hw-primary-500')
 
   function validateShortName(organization: OrganizationGeneralInfoDTO) {
-    if (organization.shortName === '') {
+    if (organization.shortName.trim() === '') {
```
MaxSchaefer, line 100: It would be smart if we move `organization.shortName` into a separate variable. This variable should be then used throughout the rest of the function. This applies for all `validate...()` functions.

MaxSchaefer (approved):

MaxSchaefer (approved):

PR #485, opened 2023-07-04
pin versions

### LINKED ISSUE #484: Pin versions
### Describe the chore

Currently, we are facing issues when running `pnpm install` throwing error, because packages are updated. Until the time, we are upgrading our package versions, we should pin them to a working one.

DasProffi:
closes #484


MaxSchaefer:
Please use https://docs.renovatebot.com/configuration-options/#rangestrategy instead of manually pinning the dependencies.

MaxSchaefer (changes requested):
Please use https://docs.renovatebot.com/configuration-options/#rangestrategy instead of manually pinning the dependencies.

use-to:
Yeah, this is pain, but shouldn't we just fix the issues after installing new packages instead?

DasProffi:
> Yeah, this is pain, but shouldn't we just fix the issues after installing new packages instead?

That's the plan, but with the current changes to information passing with contexts and the backend integration, changing the code after a major update of the packages would be too much workload to make effective progress.

Therefore, we want to delay this for now by version pinning

use-to:
This pr has currently no progress. Is help wanted @DasProffi ?

DasProffi:
> This pr currently has no progress. Is help wanted @DasProffi ?

I discussed it with @MaxSchaefer he wanted to check my proposed versioning and then change the renovate config

MaxSchaefer (approved):

PR #513, opened 2023-07-16
add manage beds modal and use it in roomlist

### LINKED ISSUE #502: Manage beds modal
### Describe the feature

The manage beds modal should list all beds inside a room via cards. All bed cards have a "Remove" button. This button is greyed out when the bed is occupied by a patient. A new bed can be created inside the room via a placeholder card "Create bed".

The current management of beds inside are room are very limited with the number input field and not very intuitive. For example, a removal of a bed would take place via a decrementation of the bed counter. This is not very explicit.

### Additional context

Tasks - overview of wards in an organization (1)
Tasks - overview of wards in an organization (2)

DasProffi:
closes #502

### DIFF
--- a/lib/util/noop.ts
+++ b/lib/util/noop.ts
@@ -1,0 +1,1 @@
+export const noop = () => undefined
--- a/tasks/mutations/bed_mutations.ts
+++ b/tasks/mutations/bed_mutations.ts
@@ -1,4 +1,14 @@
 import type { PatientDTO, PatientWithTasksNumberDTO } from './patient_mutations'
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
+import { bedService, getAuthenticatedGrpcMetadata } from '../utils/grpc'
+import {
+  CreateBedRequest, DeleteBedRequest,
+  GetBedRequest,
+  UpdateBedRequest
+} from '@helpwave/proto-ts/proto/services/task_svc/v1/bed_svc_pb'
+import { roomOverviewsQueryKey, roomsQueryKey } from './room_mutations'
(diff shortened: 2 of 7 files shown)


lib/util/noop.ts
MaxSchaefer, line 1: 👀

lib/util/noop.ts
```diff
@@ -0,0 +1 @@
+export const noop = () => undefined
```
MaxSchaefer, line 1: 👀

tasks/mutations/bed_mutations.ts
```diff
@@ -29,4 +39,107 @@ export type BedMinimalDTO = {
...
+
+export const useBedQuery = (bedID: string | undefined) => {
+  return useQuery({
+    queryKey: [bedQueryKey],
+    enabled: !!bedID,
+    queryFn: async () => {
+      const req = new GetBedRequest()
+      if (bedID) {
+        req.setId(bedID)
+      }
```
MaxSchaefer, line 57: A ID is required by the API. We should early return when `bedID` is falsy

MaxSchaefer (changes requested):

MaxSchaefer:
@DasProffi please scale the table to the width of the modal.
Bildschirm­foto 2023-07-18 um 12 30 32 AM

tasks/mutations/bed_mutations.ts
DasProffi: The query isn't excited when the bed Id isn't defined, see the enabled attribute. The if only occurs because Typescript doesn't see that.

MaxSchaefer:
Awesome! Now we can properly manage beds 🚀

PR #525, opened 2023-07-25
force unchecked state for task template subtasks

### LINKED ISSUE #519: Checkboxes are mutable on task templates
### Describe the bug

The checkboxes for subtasks by editing task templates are mutable.
Obviously, these subtasks will be unchecked when using the template on a real task.
So it makes no differences in weather the subtask is checked or not.

We should make the checkboxes for subtasks unchecked the entire time and disable them, too.

Just edit any task template and add subtasks to it.

### Additional context

#505

DasProffi:
closes #519

### DIFF
--- a/tasks/components/SubtaskView.tsx
+++ b/tasks/components/SubtaskView.tsx
@@ -101,10 +101,10 @@ export const SubtaskView = ({
                 }}
                 onDoneChange={done => {
                   // taskTemplateId === "" for the creation of a template
-                   if (taskTemplateId) {
+                  if (taskTemplateId) {
                     return
                   }
-                  
+
                   subtask.isDone = done
                 }}
               />
--- a/tasks/components/SubtaskView.tsx
+++ b/tasks/components/SubtaskView.tsx
@@ -100,7 +100,10 @@ export const SubtaskView = ({
                   }
                 }}
                 onDoneChange={done => {
-                  subtask.isDone = done
+                  // taskTemplateId === "" for the creation of a template
+                  if (taskTemplateId === undefined) {
+                    subtask.isDone = done
+                  }
                 }}
               />
             ))}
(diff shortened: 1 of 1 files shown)


use-to (changes requested):

florianjs:
We should also consider the case when a new subtask is created, then the ID is also empty and we should check for undefindend.

tasks/components/SubtaskView.tsx
```diff
@@ -100,6 +100,11 @@ export const SubtaskView = ({
                   }
                 }}
                 onDoneChange={done => {
+                  // taskTemplateId === "" for the creation of a template
```
florianjs, line 103: if (taskTemplateId || taskTemplateId === '') {
     return
 }

florianjs (changes requested):

tasks/components/SubtaskView.tsx
```diff
@@ -100,6 +100,11 @@ export const SubtaskView = ({
...
+                  if (taskTemplateId) {
```
florianjs, line 104:
```suggestion
    if (taskTemplateId || taskTemplateId === '') {
      return
    }
```

use-to (approved):

tasks/components/SubtaskView.tsx
DasProffi: creates are should indicated by an empty string for id and not undefined

florianjs: Yes, but then we still have the problem that the problem occurs when creating the checkbox.

florianjs: > creates should be indicated by an empty string for id and not undefined

Yes, but we should check of !== undefinded because if we have an empty string it's not undefinded

PR #537, opened 2023-07-27
use revoke mutation

### LINKED ISSUE #533: Use Revoke Invitation in Organization Invite List
### Describe the feature

Use Revoke Invitation in Organization Invite List

Because decline invitation doesn't work

### Additional context

_No response_

DasProffi:
closes #533

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -199,8 +199,8 @@ importers:
         specifier: workspace:*
         version: link:../lib
       '@helpwave/proto-ts':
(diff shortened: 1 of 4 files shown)


DasProffi:
Doesn't work in the test data, because the test user isn't an admin of an organization and therefore doesn't has permission to use the revoke invitation endpoint. The Remove isn't disabled, because no proper method for checking the admin status of a user exists yet

use-to (approved):

PR #585, opened 2023-08-05
TaskDetails being overwritten

### LINKED ISSUE #569: Old Task Data in new Task
### Describe the bug

Data from an delete Task is shown when creating a new on

Delete a Task and The create a new one. The data of the task befor is still there.
also the same bug as in #568 

### Additional context

grafik

DasProffi:
closes #569

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,9 +1,5 @@
 lockfileVersion: '6.0'
 
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
-
 importers:
 
   .: {}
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
florianjs, line 1: Do we really need the pnpm-lock.yml

DasProffi: As an application should contain it's lock file and the contents changed due to the new protobuf, the new lock file should be uploaded as well.
Or do you have a reason why it shouldn't be?

MaxSchaefer (approved):

PR #630, opened 2023-08-15
reenable isPublic for tasks and add Modal Stacking

### LINKED ISSUE #558: Visibility for task
### Describe the feature

Re-Enable the public/private-flag logic again and sync the field with the API.

The user needs to set private and public tasks.

### Additional context

[The logic was disabled by this pr.](https://github.com/helpwave/web/pull/520)

### LINKED ISSUE #625: Modal Stacking
### Describe the feature

Use a Context to register modals via an id and update/display them when appropriate an also allow for deregistering

Our current Modal allows for Stacking multiple modals on top of each other, but the background get dimmed with every other modal. Only the top layered/most recent one should have a shadow.

### Additional context

_No response_

DasProffi:
closes #558
closes #625

~The backend isn't working properly on create see https://github.com/helpwave/services/pull/359~

For the modal stacking look at the storybook

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,9 @@ This repository is split up into multiple subprojects using [pnpm](https://pnpm.
 1. open [`http://localhost:3000`](http://localhost:3000)
 
 ## Storybook
+`pnpm run storybook` in the lib folder
 
-We are about to adopting Storybook to develop, preview, testing and showcasing our components. Below you will find a listing of our deployt Storybook. You can also run these locally to develop new stories via `pnpm run storybook` (open the displayed port in your browser of choice) in the supported projects.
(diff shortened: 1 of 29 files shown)


MaxSchaefer (changes requested):
Very nice! To points:

- [ ] A "you cannot undo this operation"-modal before going public with a task
- [ ] Re-enable the red outline of a task that is already due

MaxSchaefer (commented):
I like the modal. @DasProffi after I confirmed  the modal, the tasks gets only published when I save/update the task. The UX is not quite nice. As a user I would expect that the task gets published directly after my confirmation. If this needs a large refactor, let me know.

DasProffi:
> I like the modal. @DasProffi after I confirmed the modal, the tasks gets only published when I save/update the task. The UX is not quite nice. As a user I would expect that the task gets published directly after my confirmation. If this needs a large refactor, let me know.

Yes, that's an issue for the entire task screen, see #636

MaxSchaefer (approved):

PR #646, opened 2023-08-19
update feature display and fix TwoColumn using the wrong width

### LINKED ISSUE #645: Feature list wide layout no padding
### Describe the bug

there is no padding between the title and the picture
image

Change the slider on the index page

### Additional context

_No response_

DasProffi:
closes #645

### DIFF
--- a/tasks/components/layout/TwoColumn.tsx
+++ b/tasks/components/layout/TwoColumn.tsx
@@ -78,6 +78,7 @@ export const TwoColumn = ({
   }
 
   // TODO Update this to be more clear and use all/better constraints
+  // LINTER: needs this anonymous function to be happy :(
   const calcPosition = (() => (dragPosition: number) => {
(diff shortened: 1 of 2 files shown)


tasks/components/layout/TwoColumn.tsx
```diff
@@ -141,7 +145,7 @@ export const TwoColumn = ({
             className={tw(`overflow-hidden`)}
             style={{ width: leftWidth + 'px' }}
           >
-            <SimpleBarReact ref={scrollableRefLeft} style={{ overflowX: 'hidden', maxHeight: simpleBarMaxHeight }}>
```
florianjs, line 144: why do you removed overflowX: 'hidden'?

DasProffi: Because it causes the width calculation to be off and with the corrected version the width should always match

use-to (approved):
Finally no more warnings...

PR #650, opened 2023-08-21
draft for internationalizing news feed

### LINKED ISSUE #647: Internationalize Feed
### Describe the chore

Feed URL containing language based content: https://cdn.helpwave.io/feed_intl.json

- [ ] So, as you can see, use the proper language based on the user settings.
- [ ] Also, please filter out feeds not containing the `"tasks"`-key inside the `keys`-field.
- [ ] Move feed logic to `lib/` and make it generic.

DasProffi:
closes #647


use-to:
@JannikWibker could you take a look at this zod mess?

lib/util/news.ts
```diff
@@ -0,0 +1,43 @@
+import type { Languages } from '../hooks/useLanguage'
+import { languages } from '../hooks/useLanguage'
+import { z } from 'zod'
+
+export type News = {
+  title: string,
+  date: Date,
+  description: (string | URL)[],
```
JannikWibker, line 8: Just a thought but you may want to allow (a subset of) markdown here

I'd say "inline formatting" such as bold, italic, strikethrough, links, inline code, as well as lists sound like a reasonable subset

I'm not sure if there is some special nextjs ecosystem solution for this which allows such a fine-grained level of control but you could always try to use markdown-it or some other markdown parser which tend to be more configurable on that front

lib/util/news.ts
```diff
@@ -0,0 +1,43 @@
...
+  externalResource?: URL,
+  keys: string[]
+}
+
+export type LocalizedNews = Record<Languages, News[]>
+
+export const newsSchema = z.object({
+  title: z.string(),
+  description: z.string(),
+  date: z.string(),
```
JannikWibker, line 18: looking at http://cdn.helpwave.de/feed_intl.json shows this to be a string such as `"10 July 2023"`, I'd suggest sticking to ISO 8601 even though you might end up with some timezone trickery. Do note that currently your dates are not localized per language but all in english.

You can use the Intl api for displaying the dates: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

lib/util/news.ts
```diff
@@ -0,0 +1,43 @@
...
+  keys: z.array(z.string())
+}).transform<News>((obj) => {
+  let description: (string | URL)[] = [obj.description]
+  if (obj.image) {
+    description = [new URL(obj.image), ...description]
+  }
+
+  return {
+    title: obj.title,
+    date: new Date(obj.date),
```
JannikWibker, line 30: what the date constructor can parse may differ between browsers. The spec doesn't state all too much in this regard (afaik) and many browsers just started to do some kind of "best effort" strategy and started supporting many many different formats, thus this is kind of a mess and sticking to ISO 8601 would definitely be the most safe way forward

tasks/components/layout/NewsFeed.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { filterNews } from '@helpwave/common/util/news'
+
+type NewsFeedTranslation = {
+  title: string,
+  noNews: string
+}
+
+const defaultNewsFeedTranslations: Record<Languages, NewsFeedTranslation> = {
+  en: {
+    title: 'What\'s new in helpwave tasks?',
```
JannikWibker, line 18: use `"What's new in helpwave tasks?"` here (no backslash escape)

tasks/components/layout/NewsFeed.tsx
```diff
@@ -0,0 +1,62 @@
...
+ */
+export const NewsFeed = ({
+  language,
+  localizedNews,
+  width
+}: PropsWithLanguage<NewsFeedTranslation, NewsFeedProps>) => {
+  const translation = useTranslation(language, defaultNewsFeedTranslations)
+  // The value of how much space a FeatureDisplay needs before the title can be displayed on its left
+  // Given in px
+  const widthForAppearanceChange = 600
```
JannikWibker, line 43: `WIDTH_FOR_APPEARANCE_CHANGE`

tasks/utils/news.ts
```diff
@@ -0,0 +1,14 @@
...
+import type { LocalizedNews } from '@helpwave/common/util/news'
+import { localizedNewsSchema } from '@helpwave/common/util/news'
+
+export const fetchLocalizedNews = (): Promise<LocalizedNews> => {
+  // TODO change later back
+  const { featuresFeedUrl } = { featuresFeedUrl: 'https://cdn.helpwave.io/feed_intl.json' }// getConfig()
+  return fetch(featuresFeedUrl)
+    .then((res) => res.json())
+    .then(async (json) => {
+      await localizedNewsSchema.parseAsync(json)
```
JannikWibker, line 11: why are you using `.parseAsync` here?

this could just be the following:

```ts
.then((json) => {
  const maybeLocalizedNews = localizedNewsSchema.safeParse(json)
  if (!maybeLocalizedNews.success) {
    // TODO: throw custom error with a useful message
  } else {
    return maybeLocalizedNews.data
  }
})
```

You should return `maybeLocalizedNews.data` here as this has the transforms applied

tasks/pages/index.tsx
```diff
@@ -66,9 +68,10 @@ const Dashboard: NextPage<PropsWithLanguage<DashboardTranslation, DashboardServe
             )
             )}
             right={width => (
-              <FeatureDetails
+              <NewsFeed
                 width={width}
-                features={featuresSchema.parse(jsonFeed)}
+                // TODO fix typing
+                localizedNews={localizedNewsSchema.parse(jsonFeed) as LocalizedNews}
```
JannikWibker, line 74: Why are you typechecking and parsing this here again?, `fetchLocalizedNews` already does this

tasks/utils/news.ts
MaxSchaefer: Due to the fact that `getServerSideProps` wants to return a serializable struct. This was not possible with the date inside. I don't know if `.date` solves that, this would be great, and we don't have to reparse this again.

JannikWibker: Oh I see, didn't consider this. In that case this makes sense

`.date` wouldn't solve this

MaxSchaefer: Yes. I've tried this out right now .. :)
```
Error: Error serializing `.localizedNews.de[0].date` returned from `getServerSideProps` in "/".
Reason: `object` ("[object Date]") cannot be serialized as JSON. Please only return JSON serializable data types.
```

tasks/pages/index.tsx
MaxSchaefer: _See below_

@JannikWibker what would you suggest to get rid of this `as ...`?

tasks/utils/news.ts
```diff
@@ -0,0 +1,14 @@
+// import { getConfig } from './config'
+import type { LocalizedNews } from '@helpwave/common/util/news'
+import { localizedNewsSchema } from '@helpwave/common/util/news'
+
+export const fetchLocalizedNews = (): Promise<LocalizedNews> => {
+  // TODO change later back
+  const { featuresFeedUrl } = { featuresFeedUrl: 'https://cdn.helpwave.io/feed_intl.json' }// getConfig()
```
use-to, line 7: Use `feed.json` as before.

use-to (changes requested):

use-to (approved):

PR #675, opened 2023-08-31
feat: update queries update behavior

### LINKED ISSUE #673: Updating ward does not update ward card
### Describe the bug

image

As you can see on the right I added a room with some beds (beds doesn't matter in this case as a room could be empty) and the left sided ward card still shows 0 beds.

1. Create a ward and edit that exact one (or edit anyone)
2. Add a room (with or without beds)
3. See on the left that the bed count (indicated by the icon) is not updating

### Additional context

The solution should be simple: Just invalidate the react query key in order to invoke a refetch afterwards.

DasProffi:
closes #673


MaxSchaefer:
@DasProffi any progress here? You could merge main for fixing the preview.

use-to (approved):

PR #682, opened 2023-09-01
add refetch interval to queries

### LINKED ISSUE #679: Continuously refetching using `refetchInterval` on `useQuery()`
### Describe the feature

We could set a default `refetchInterval` for every `useQuery()` hook via the [Default Query Function](https://tanstack.com/query/v4/docs/react/guides/default-query-function). The passed `refetchInterval` should be configurable via the config and defaults to five seconds.

[`refetchInterval` on `useQuery()`](https://tanstack.com/query/v4/docs/react/reference/useQuery)

To enable more real-time collaboration in helpwave tasks we should update / refetch the state more often. React Query can help use with that by automatically refetching after a specific time.

### Additional context

_No response_

DasProffi:
closes #679

Load the index and look at the network tab of your browser

## NOTES
Some of our components are behaving confusingly with this new refetching. As refetches are overwriting unsaved changes or are causing the changes to be saved when the save interval runs out.


MaxSchaefer:
> Some of our components are behaving confusingly with this new refetching. As refetches are overwriting unsaved changes or are causing the changes to be saved when the save interval runs out.

Makes sense. How should we handle this? Disable the refetch for specific queries on edit?

DasProffi:
> > Some of our components are behaving confusingly with this new refetching. As refetches are overwriting unsaved changes or are causing the changes to be saved when the save interval runs out.
> 
> Makes sense. How should we handle this? Disable the refetch for specific queries on edit?

Maybe, or we just disable the updates while editing? Because after a completed edit, there will be an update

MaxSchaefer:
> Maybe, or we just disable the updates while editing? Because after a completed edit, there will be an update

Sounds good. Could you propose a generic solution for disabling the updates while editing?

DasProffi:
> > Maybe, or we just disable the updates while editing? Because after a completed edit, there will be an update
> 
> Sounds good. Could you propose a generic solution for disabling the updates while editing?

Not a generic solution, but we can use a flag to indicate edits and only after an edit unset it to load/update with the request data

DasProffi:
Simple edit protection was added, yet there are some cases where updates happen slowly/need some seconds to show.

A better update management would require a refactoring of forms, which hasn't happened yet

use-to:
@MaxSchaefer Disabling autosave or using the interval polling? With this pr scope, we can't satisfy both.

MaxSchaefer:
Due to the fact that the auto-updating is not consistent across all forms, we are going to remove the auto-updating. @DasProffi

DasProffi:
This PR is stale for to long and any changes would need to be fully rewritten, therefore it will be closed

PR #684, opened 2023-09-01
switch to react custom scrollbars

### LINKED ISSUE #670: switcht to react-custom-scrollbars
### Describe the feature

we should switch from simplebar-react to react-custom-scrollbars

This feature is needed because of some problems occurred by simplar-react

### Additional context

_No response_

florianjs:
closes #670


lib/components/Table.tsx
```diff
@@ -232,7 +231,6 @@ export const Table = <T, >({
   return (
     <div className={tw('flex flex-col gap-y-4 overflow-hidden')}>
       <div>
-        <SimpleBarReact>
```
DasProffi, line 235: Why did you remove the Scrollbar wrapper and not replace it? Does the new scrollbar appear automatically?

DasProffi (changes requested):

DasProffi:
The scrollbar for the table causes layout problems and for some reason the members table is missing completely
image

florianjs:
> The scrollbar for the table causes layout problems and for some reason the members table is missing completely image

Thanks, the problem is if enable autoheigth without a min height, then the whole table is scrollable

lib/components/Table.tsx
florianjs: Not sure if we need it really because we have the pagination, so there should be the need of scrolling. The Problem is react-custom-scrollbars-2 come with some side effects, on solution I found is to remove it or to use simplebar

florianjs: [Uploading Bildschirmaufzeichnung vom 2023-09-02, 12-51-29.webm…]()

DasProffi (dismissed):

DasProffi (approved):

PR #687, opened 2023-09-01
update select options to use custom labels

### LINKED ISSUE #683: Select Component custom labels
### Describe the feature

The Select label should also accept React Nodes and not just strings

Because we might want to use an image in a Select or highlight some text

### Additional context

_No response_

DasProffi:
closes #683

Look at the new storybook for the select


MaxSchaefer (approved):
Uii nice :) We could also add an example with images

PR #751, opened 2023-10-05
add SearchableSelect to task AssigneeSelect

### LINKED ISSUE #741: Searchable select for task assignee
### Describe the feature

When the user opens the dropdown, the user should be able to search through the list by start typing. Please develop this component generic.

The user should be able to search through the list of possible assignees for faster assignment. 

### Additional context

_No response_

DasProffi:
closes #741

### DIFF
--- a/lib/components/user_input/Input.tsx
+++ b/lib/components/user_input/Input.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import type { HTMLInputTypeAttribute, InputHTMLAttributes } from 'react'
 import { tw, tx } from '../../twind'
 import { Span } from '../Span'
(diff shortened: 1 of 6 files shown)


use-to (changes requested):
Looks good to me, but two more ideas:
1. could we autofocus the input field on select click?
2. we should limit the displayed organization members to at maximum 3 or 5

DasProffi:
> 1. could we autofocus the input field on select click?

Will be done
> 2. we should limit the displayed organization members to at maximum 3 or 5

Only display 5 in a scrollable or actually only show 5?

use-to:
> > 1. could we autofocus the input field on select click?
> 
> Will be done
> 
> > 2. we should limit the displayed organization members to at maximum 3 or 5
> 
> Only display 5 in a scrollable or actually only show 5?

Scrollable would be nice, too. If we want to implement lazy loading later on we could do that.
Feel free!

DasProffi:
> Scrollable would be nice, too. If we want to implement lazy loading later on we could do that. Feel free!

The default `Select`, already scrolls, if the content exceeds more than 30% of the screen height, thus this is already implemented

use-to (approved):
Very nice! You're right, the selection is scrollable. But let's keep this in mind for lazy loading later on...

@MaxSchaefer The tasks preview on fly.io is not updating... Maybe this is due to the cancellation before.

PR #818, opened 2023-12-26
update eslint setup

### LINKED ISSUE #817: Improve eslint setup
### Describe the chore

Right now the eslint setup revolves around a single config file which can be found in the root of the repo. It isn't completely clear how (i.e. how the config file is found) and where linting is done (`lib/` has no `lint` command for example) and including the nextjs defaults isn't optimal for sub packages which don't use nextjs.

Creating a separate sub-package, which contains only an eslint config and extending this in each of the 3 sub-packages is a possible solution to this, this also allows more fine-grained control over things such as the nextjs defaults:

```js
module.exports = {
  extends: [
    '@helpwave/eslint-config'
  ]
}
```

vs.

```js
module.exports = {
  extends: [
    '@helpwave/eslint-config',
    'plugin:@next/next/recommended'
  ]
}

```

JannikWibker:
closes #817


tasks/tsconfig.json
```diff
@@ -1,25 +1,43 @@
...
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true
+    "incremental": true,
+    "forceConsistentCasingInFileNames": true,
+    "allowSyntheticDefaultImports": true,
+    // "noPropertyAccessFromIndexSignature": true, // TODO: reenable later (didn't know that tsconfig.json supported comments lol)
```
MaxSchaefer, line 22: `noPropertyAccessFromIndexSignature` sounds good! When is later?

MaxSchaefer (changes requested):
Linting of `lib` 🎉 via a centralized config! Very nice, thank you.

tasks/components/RoomOverview.tsx
```diff
@@ -80,6 +79,7 @@ export const RoomOverview = ({ room }: RoomOverviewProps) => {
               </Droppable>
             ) : (
               // Maybe also wrap inside the drag and drop later
+              // TODO: could we use a generic parameter here somewhere to get rid of the AnyData type?
```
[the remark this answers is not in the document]
JannikWibker: I am working on a general clean up PR at the same time; I'll create an issue for that and deal with the type issue there

tasks/tsconfig.json
JannikWibker: In a follow up clean up PR, don't want to change even more in a single PR 😅 
There are quite a few places where this causes errors, I'll try to fix them and then turn on the setting. Currently in the process of doing a lot of clean up already; will see if I can reasonably fit these changes in with that PR

MaxSchaefer (approved):

PR #826, opened 2023-12-27
Feat: Cleanup

### LINKED ISSUE #793: Use right organization id for all requests
### Describe the chore

Use right organization id for all requests, currently we are defaulting to a specific id instead of using the one of the current organization

### LINKED ISSUE #825: Get rid of empty strings for ids (and other bad type and state handlings)
### Describe the bug

There are a bunch of places where empty strings are used instead of `undefined` just to conform to a more stricter type. This isn't a fix but rather duct tape on top of otherwise shaky type and state management

Search for: `id: ''` or `?? ''` to see for yourself, there are a few places where this is fine, but not to this extend

### Additional context

_No response_

JannikWibker:

Closes #825 and probably a few more issues I'm about to create
Maybe closes #793, at least one occurrence of this seems to be fixed by this PR

There are quite a few new TODOs with questions about state related things, it is probably easiest to just use code review comments to answer some of those


tasks/mutations/ward_mutations.ts
```diff
@@ -119,28 +119,26 @@ export const useWardDetailsQuery = (wardId?: string, organisationId?: string) =>
...
+    const res = await wardService.getWard(req, getAuthenticatedGrpcMetadata(organisationId))
 
-      const ward: WardWithOrganizationIdDTO = {
-        id: res.getId(),
-        name: res.getName(),
-        organizationId: res.getOrganizationId()
-      }
-      return ward
+    if (!res.toObject()) {
+      console.error('error in Ward query')
```
use-to, line 131: throw error? Haven't you applied that to every `console.error` pattern?

tasks/pages/templates.tsx
```diff
@@ -78,15 +78,15 @@ export const TaskTemplateContext = createContext<TaskTemplateContextType>({
...
-  const { templateId } = router.query
+  const templateId = useRouteParameters<never, 'templateId'>().templateId
   const [usedQueryParam, setUsedQueryParam] = useState(false)
   const { user } = useAuth()
   const { isLoading, isError, data } = usePersonalTaskTemplateQuery(user?.id)
 
   const [contextState, setContextState] = useState<TaskTemplateContextState>(taskTemplateContextState)
 
-  const createMutation = useCreateMutation('personalTaskTemplates', taskTemplate =>
+  const wardId = '' // TODO: why is this empty? (tracing back where is value originally came from lead me to the fact that this value would always be undefined or an empty string, thus I hardcoded it here now but it would be great to know if this is correct)
```
use-to, line 88: @DasProffi

tasks/pages/ward/[wardId].tsx
```diff
@@ -167,30 +165,30 @@ const WardOverview: NextPage = ({ language }: PropsWithLanguage<WardOverviewTran
...
+        if (target.patientListSection === 'unassigned') {
+          if ('discharged' in source && source.discharged) {
             readmitPatientMutation.mutate(patientId)
-          } else {
-            unassignMutation.mutate(patientId)
           }
-          unassignMutation.mutate(patientId)
-        } else if (overData.patientListSection === 'discharged') {
+          unassignMutation.mutate(patientId) // TODO: is this correct?, doing two mutations for a patient if previously discharged?
+        } else if (target.patientListSection === 'discharged') { // TODO: can we unify the formats for patients PREVIOUSLY having been discharged and patients BEING discharged? (boolean vs.patientListSection)
```
Comment (use-to) [inline tasks/pages/ward/[wardId].tsx:181]: @DasProffi

tasks/pages/ward/[wardId].tsx
```diff
@@ -211,18 +209,24 @@ const WardOverview: NextPage = ({ language }: PropsWithLanguage<WardOverviewTran
...
-      assignBedMutation.mutate({
-        id: contextState.bedId,
-        patientId: patient.id
+  // TOOD: can we somehow assert that the patient is not undefined here?
+  const createPatient = () => contextState.patient && createPatientMutation.mutateAsync(contextState.patient)
+    .then((patient) => {
+      // TODO: is it a good idea to have this much state in one object?, also this shouldn't be called context I think
+      setContextState({
+        ...contextState,
+        patient,
```
Comment (use-to) [inline tasks/pages/ward/[wardId].tsx:220]: @DasProffi

use-to (approved):
Large pr but excellent work. Consider refactoring `console.error`.

use-to:
@DasProffi Could you clarify the with `// TODOs` marked code lines by @JannikWibker ?

tasks/pages/templates.tsx
DasProffi: There should be no need for a ward id in this view as these are personal templates. But the ward id from the router.query is used in the TaskTemplateDisplay.tsx component

tasks/pages/ward/[wardId].tsx
Comment (DasProffi) [reply tasks/pages/ward/[wardId].tsx:181]: doing two mutations is correct as we need to reactivate a patient, before using it again

a boolean would not be good as we have three possible states active, discharged, assigned. Assigned cannot be dragged to, but this might change later on

tasks/pages/ward/[wardId].tsx
Comment (DasProffi) [reply tasks/pages/ward/[wardId].tsx:220]: we cannot assert that patient is not undefined as this indicates whether a patient will be created and if not set a patient isn't created. This might be solved by splitting the state as you suggested, but will be much effort for checking that all tests against the context stay the same.

Regarding the rename: This is the state of the provided context for children, if you prefer a different naming, I won't object

Comment (JannikWibker) [reply tasks/pages/ward/[wardId].tsx:220]: The thing with the word context is that there exists the concept or [react contexts](https://react.dev/reference/react/createContext); I wouldn't use a name that is already taken by something else in the same domain

MaxSchaefer (dismissed):

tasks/mutations/ward_mutations.ts
JannikWibker: I wanted to refactor this hook a bit more and haven't gotten around to that yet; didn't want to just partially update the code

tasks/pages/organizations.tsx
```diff
@@ -45,8 +45,7 @@ export const OrganizationContext = createContext<OrganizationContextType>({
  */
 const OrganizationsPage: NextPage = ({ language }: PropsWithLanguage<OrganizationsPageTranslation>) => {
   const translation = useTranslation(language, defaultOrganizationsPageTranslation)
-  const router = useRouter()
-  const { organizationId } = router.query
+  const organizationId = useRouteParameters<never, 'organizationId'>().organizationId
```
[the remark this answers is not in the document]
JannikWibker: Using object destructuring ignores the `noPropertyAccessFromIndexSignature ` check, thus I updated the code to not use it where ever it is easily possible

use-to:
@JannikWibker do you mind merging?

JannikWibker:
I'd kind of like to finish all things I started in this PR but as I'm really really short on time right now I guess merging is the best thing to do right now 😅 

Maybe I'll get around to doing a bit more cleanup in the near future

use-to (approved):

PR #854, opened 2024-02-08
products dropdown

### LINKED ISSUE #849: Products dropdown
### Describe the feature

Use a dropdown name products to show the products

Because we will have more than one product

### Additional context

_No response_

DasProffi:
closes #849

### DIFF
--- a/landing-page/components/Header.tsx
+++ b/landing-page/components/Header.tsx
@@ -1,13 +1,12 @@
 import { Span } from '@helpwave/common/components/Span'
 import { Helpwave } from '@helpwave/common/icons/Helpwave'
 import { tw } from '@helpwave/common/twind'
-import { Menu, X } from 'lucide-react'
+import { X, Menu as MenuIcon } from 'lucide-react'
 import { useState } from 'react'
 import Link from 'next/link'
-import { MenuItem } from '@helpwave/common/components/user-input/Menu'
-import { router } from 'next/client'
+import { Menu, MenuItem } from '@helpwave/common/components/user-input/Menu'
 
-const homeURL = '/'
+const homeURL: string = '/'
 
 type LinkType = {
   name: string,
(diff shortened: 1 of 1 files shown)


landing-page/components/Header.tsx
```diff
@@ -1,18 +1,47 @@
...
-  { name: 'team', url: '/team' },
-  { name: 'talks', url: '/talks' },
-  { name: 'contact', url: 'mailto:contact@helpwave.de' },
+type SubLinkType = LinkType & {
+  subpage?: LinkType[]
+}
+
+const items: SubLinkType[] = [
+  {
+    name: 'product',
```
use-to, line 22:
```suggestion
    name: 'products',
```

landing-page/components/Header.tsx
```diff
@@ -1,18 +1,47 @@
...
+    url: '/product',
```
use-to, line 23:
```suggestion
    url: '/products',
```

Not sure if there will be a subpage on `/products`...

use-to: Maybe this button should not be clickable as a Link.

use-to (changes requested):
Can we make it a little wider? Maybe a 2 column layout. Add some mock routes for this purpose (cloud, scaffold, core, analytics).
Also, can you trigger the dropdown on hover, too? This is more convenient on desktop devices.

use-to:
Good work, keep it up! Looks awesome even now.

DasProffi:
> Can we make it a little wider? Maybe a 2 column layout. 

What exactly are you referring to? This issue should only implement the dropdown and routing, but we can open another issue for two column screens.

> Add some mock routes for this purpose (cloud, scaffold, core, analytics). 
Also, can you trigger the dropdown on hover, too? This is more convenient on desktop devices.

Mock routes and On hover are implemented now.

use-to (approved):
Looks very nice!
I was referring to a two column drop down menu.
This would split up the menu items into two columns.
So we can add some more content and the menu is wider.
Anyway, this works for now. We can definitely slice this up into another issue.

PR #896, opened 2024-02-26
add DateTimePicker base and YearMonthPicker

### LINKED ISSUE #895: Date And Time Picker
### Describe the feature

Write or import a DateTimePicker

Because we use two different date and time pickers in the `TaskDetailView` and the `DateProperty` we currently use the browser's default date time picker which is not stylized to our demands.

### Additional context

from #887

DasProffi:
closes #895

### DIFF
--- a/lib/components/Expandable.tsx
+++ b/lib/components/Expandable.tsx
@@ -1,0 +1,43 @@
+import type { PropsWithChildren, ReactNode } from 'react'
+import { forwardRef, useState } from 'react'
+import { ChevronDown, ChevronUp } from 'lucide-react'
+import { tw, tx } from '../twind'
+
+export type ExpandableProps = PropsWithChildren<{
+  label: ReactNode,
+  icon?: (expanded: boolean) => ReactNode,
+  initialExpansion?: boolean,
+  className?: string,
+  ref?: React.Ref<HTMLDivElement>
+}>
+
+/**
+ * A Component for showing and hiding content
+ */
+export const Expandable = forwardRef<HTMLDivElement, ExpandableProps>(({
+  children,
+  label,
+  icon,
+  initialExpansion = false,
+  className = '',
+}, ref) => {
+  const [expanded, setExpanded] = useState(initialExpansion)
+  icon ??= expanded1 => expanded1 ? <ChevronUp size={16}/> : <ChevronDown size={16}/>
+
+  return (
+    <div ref={ref} className={tx('flex flex-col', className)}>
+      <div
+        className={tw('flex flex-row justify-between items-center cursor-pointer')}
+        onClick={() => setExpanded(!expanded)}
+      >
+        {label}
+        {icon(expanded)}
+      </div>
+      <div className={tx('flex flex-col', { 'h-full': expanded, 'h-0 overflow-hidden': !expanded })}>
+        {children}
+      </div>
+    </div>
+  )
+})
+
+Expandable.displayName = 'Expandable'
--- a/lib/components/Stories/DateTimePicker.stories.tsx
+++ b/lib/components/Stories/DateTimePicker.stories.tsx
@@ -1,0 +1,20 @@
+import type { Meta, StoryObj } from '@storybook/react'
+import { DateTimePicker } from '../user-input/DateAndTimePicker'
+import { addDuration, subtractDuration } from '../../util/date'
+
+const meta = {
+  title: 'Input',
+  component: DateTimePicker,
+} satisfies Meta<typeof DateTimePicker>
+
+export default meta
+type Story = StoryObj<typeof meta>;
+
+export const DateTimePickerVariation: Story = {
+  args: {
+    mode: 'dateTime',
+    value: new Date(),
+    start: subtractDuration(new Date(), { years: 50 }),
+    end: addDuration(new Date(), { years: 50 }),
+  },
+}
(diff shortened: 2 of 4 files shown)


lib/components/user-input/ScrollPicker.tsx
JannikWibker, line 1: Right now if you don't stop scrolling at the exact spot the input jumps around, use `scroll-snap-type` to make this smoother (the scrolling itself will smoothly snap to a target, with this the need for updating the scroll position manually should go away)

- https://tailwindcss.com/docs/scroll-snap-type
- https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type

lib/components/date/DatePicker.tsx
```diff
@@ -0,0 +1,156 @@
...
+  onChange?: (date: Date) => void,
+  dayPickerProps?: Omit<DayPickerProps, 'value'|'onChange'|'selected'>,
+  yearMonthPickerProps?: Omit<YearMonthPickerProps, 'value'|'onChange'|'startYear'|'endYear'>,
+  className?: string
+}
+
+/**
+ * A Component for picking a date
+ */
+export const DatePicker = ({
```
JannikWibker, line 85: image

I'd make the `26` until `29` a bit lighter

lib/components/date/DatePicker.tsx
JannikWibker, line 1: https://github.com/helpwave/web/assets/16372321/f140ded3-f0ab-4e3b-a377-27a73888b3cd

this shouldn't jump around (i.e. change its height)

JannikWibker (commented):
Generally looks good (a few minor comments though), but I'm kind of wondering why you went with implementing everything yourself rather than using existing solutions such as:
- https://github.com/Hacker0x01/react-datepicker
- https://react-spectrum.adobe.com/react-aria/useDatePicker.html
- https://github.com/aliakbarazizi/headless-datepicker

Time and date things are notoriously hard to get right and thus opting for something existing and customizing it to your liking is probably preferred.

Might want to replace some of the underlying parts of this in the future with some of these listed solutions while keeping the current styling

The current components are not keyboard accessible at all, might want to add that (but this is one of these "scope creeps" that libraries would most likely solve)

lib/components/date/DatePicker.tsx
JannikWibker: One more thing: I feel like the hover boxes should all have equal size (thus not depending on the width of the text)
It might also make sense to use a monospace font for the numbers, i cannot really tell right now if this is a good idea or not but worth a shot

lib/components/user-input/ScrollPicker.tsx
DasProffi: Will be done in #906

lib/components/date/DatePicker.tsx
DasProffi: All boxes have equal size it just does not appear that way because of the font and number of digits, using a monospace font and padded 0s could be a solution in the future

MaxSchaefer (commented):
fyi: The JavaScript Intl API could also provide a list for translated days or moths. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

JannikWibker:
Not sure wether or not this is intended but the rework of `useTranslation` removed one core feature:
- being able to quickly and effortlessly override the translation of a component

I'm not sure if this is needed or not, but the idea behind it was that this could be used as a common interface to for example replace placeholders such as "Search.." with something more specific for the exact use case without having to add potentially even multiple separate props for that

If this is not required it is fine to remove it, but I still wanted to point this out

**Sidenote**: you're mixing up `local` and `locale`, it should be called `useLocale`

DasProffi:
> Not sure wether or not this is intended but the rework of useTranslation removed one core feature:
>
> * being able to quickly and effortlessly override the translation of a component

That was intentional, as we currently don't use it in any component and it creates an overhead for all components using a Translation. I believe that a single overwrite for specific translations is more appropriate for now. Apart from that I also think that the language attribute should only hold the language and not the translation overwrite values. Maybe renaming an retyping it to something like this would be a solution that is easier to understand:
```typescript
translationOverwrite: {language: Languages, overwrite: Partial<Translation>}
```

> Sidenote: you're mixing up local and locale, it should be called useLocale

You are right, thank you

MaxSchaefer:
I also noticed this change and will dig into this later on. Regardless of whether we accept this change of useTranslation or not, it must be cherry-picked over into a new branch/pull-request. This is not a part of #895.

DasProffi:
> I also noticed this change and will dig into this later on. Regardless of whether we accept this change of useTranslation or not, it must be cherry-picked over into a new branch/pull-request. This is not a part of #895.

I moved it to a new issue and it will be implemented in #909

MaxSchaefer:
@JannikWibker could you give this PR a code review?

lib/components/date/DayPicker.tsx
```diff
@@ -0,0 +1,81 @@
...
+  while (currentDate.getDay() !== weekStartIndex) {
+    currentDate = subtractDuration(currentDate, { days: 1 })
+  }
+
+  while (currentDate.getMonth() !== (month + 1) % 12 || currentDate.getDay() !== weekStartIndex) {
+    const date = new Date(currentDate)
+    date.setHours(value.getHours(), value.getMinutes()) // To make sure we are not overwriting the time
+    dayList.push(date)
+    currentDate = addDuration(currentDate, { days: 1 })
+  }
```
JannikWibker, line 44: I think this could be extracted into a helper function

JannikWibker (commented):
Looks good to me

MaxSchaefer (approved):

DasProffi:
@MaxSchaefer The second build of the pages for Cloudflare in the CI occurs here

PR #913, opened 2024-03-07
feat: simplify translations

### LINKED ISSUE #909: simplify translations
### Describe the chore

The current Translations are using `PropsWithLanguage` which uses apart from a language attribute also a overwriting translation on that language attribute which is not intuitive and not easy to understand. Thefore this should be simplified and the overwrite moved somewhere different

DasProffi:
closes #909


DasProffi:
@MaxSchaefer would you agree that the overwrite should be a optional 2nd attribute in the `useTranslation` hook?

Currently it is the first which makes it weird when omiting it like in

tasks/components/layout/NewsFeed.tsx
```diff
@@ -31,15 +31,16 @@ export type NewsFeedProps = {
...
-}: PropsWithLanguage<NewsFeedTranslation, NewsFeedProps>) => {
-  const translation = useTranslation(language, defaultNewsFeedTranslations)
+}: PropsForTranslation<NewsFeedTranslation, NewsFeedProps>) => {
+  const translation = useTranslation(defaultNewsFeedTranslations, overwriteTranslation)
   // The value of how much space a FeatureDisplay needs before the title can be displayed on its left
   // Given in px
   const widthForAppearanceChange = 600
-  const usedLanguage = useLanguage().language
+  let usedLanguage = useLanguage().language
+  usedLanguage = overwriteTranslation?.language ?? usedLanguage
```
MaxSchaefer, line 43: Quite interesting ... we are always need to do this check if we are calling `useLanguage()` directly

MaxSchaefer (approved):

tasks/components/layout/NewsFeed.tsx
DasProffi: We only have to do this if we want to overwrite the language and want to use it. `useLanguage()` always returns a value, but to avoid such statements we could add a overwrite to useLanguage which always returns that value if it is set.

```typescript
useLanguage(overwrite?: Languages) {
  ...
  if(!!overwrite} {
    return overwrite
  }
  ...
}
```

PR #931, opened 2024-03-10
create property form

### LINKED ISSUE #928: Create property form
If the user clicks on the `Add property` button, the right column displays the "New property" form. The formular contains input groups. The first input group is opened by default. The "Subject type" select should be disabled and pre-selected by the subject type from the left-side. To proceed to the next input group, the user can click on "Next step", the opposite goes for "Back". When there is no input-group left, show a confirmation button "Create" button. We could display the upcoming property name after "New Property" as "New Property ${property.name}".

Image

DasProffi:
closes #928


MaxSchaefer (changes requested):
- During creation, the InputGroups should only be expandable via the stepper
- The last step, as a an overview is very nice! Last step: All InputGroups are expanded, not clickable by user
-  > The "Subject type" select should be disabled and pre-selected by the subject 
    type from the left-side.
- (In the stepper, the active step is highlighted. Could we do the same for the InputGroup titles? Highlighted name ("Basic Information", "Field", "Rules") when the step is active.)

DasProffi:
> The "Subject type" select should be disabled and pre-selected by the subject
>  type from the left-side.

It seems unintuitive for the user to have to first select a subject type on the opposite screen half to create a property for that type. Do we maybe only want to do a preselect for that type instead of locking it?

MaxSchaefer:
> > The "Subject type" select should be disabled and pre-selected by the subject
> > type from the left-side.
> 
> It seems unintuitive for the user to have to first select a subject type on the opposite screen half to create a property for that type. Do we maybe only want to do a preselect for that type instead of locking it?

Yes, just pre-selecting from the left-side while still be able to change the selection is fine.

tasks/pages/properties.tsx
```diff
@@ -45,13 +49,14 @@ export const PropertyContext = createContext<PropertiesContextType>({
  */
 const PropertiesPage: NextPage = ({ overwriteTranslation }: PropsForTranslation<OrganizationsPageTranslation>) => {
   const translation = useTranslation(defaultOrganizationsPageTranslation, overwriteTranslation)
-  const propertyId = useRouteParameters<never, 'id'>().id
+  const { id: propertyId, subject: subjectType } = useRouteParameters<never, 'id' | 'subject'>()
```
MaxSchaefer, line 52: I cannot set these via the params
https://pr-932-helpwave-web-tasks.fly.dev/properties?subject=patient

MaxSchaefer (approved):

tasks/pages/properties.tsx
DasProffi: You have the wrong branch id, it should be [https://pr-931-helpwave-web-tasks.fly.dev/properties?subject=patient](https://pr-931-helpwave-web-tasks.fly.dev/properties?subject=patient)

lib/components/StepperBar.tsx
```diff
@@ -0,0 +1,125 @@
...
+  onChange: (step: StepperInformation) => void,
+  onFinish: () => void,
+  showDots?: boolean,
+  className?: string
+}
+
+/**
+ * A Component for stepping
+ */
+export const StepperBar = ({
```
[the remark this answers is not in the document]
DasProffi: Will be handled in #937

PR #945, opened 2024-03-26
date property without time

### LINKED ISSUE #941: Date property without time
### Describe the feature

The [existing date property](http://localhost:6006/?path=/story/user-action-property--date-property-variation) should also have the option to input/select date only (without the time). This could either a new property or a option on the existing one. I guess the first option would probably better suited for the composition of the properties.

As an already discussed feature. services already implemented this type.

### Additional context

_No response_

DasProffi:
closes #941


lib/util/date.tsx
```diff
@@ -6,10 +6,17 @@ export type Month = typeof monthsList[number]
 export const weekDayList = ['sunday', 'monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday'] as const
 export type WeekDay = typeof weekDayList[number]
 
-export const formatDate = (date: Date) => {
+export const formatDateTime = (date: Date) => {
```
MaxSchaefer, line 9: You can call `formatDate()` here as well and apply the time string logic afterwards

lib/components/user-input/Input.tsx
```diff
@@ -26,6 +26,7 @@ export type InputProps = {
   onChangeEvent?: (event: ChangeEvent<HTMLInputElement>) => void,
   className?: string,
   onEditCompleted?: (text: string) => void,
+  onEditCompletedEvent?: (event: ChangeEvent<HTMLInputElement>) => void,
```
MaxSchaefer, line 29:
```suggestion
  onEditCompleted?: (text: string, event: ChangeEvent<HTMLInputElement>) => void,
```
Same goes for `onChangeEvent` a few lines above. I guess we could quickly refactor this as well.

MaxSchaefer (changes requested):

MaxSchaefer (approved):

PR #957, opened 2024-03-28
remove unnecessary update button on task details

### LINKED ISSUE #782: Updating tasks does not close pop-up window
### Describe the bug

When updating a task, i.e. after completing a sub task, the pop-up window is not closed. This is opposed to the behavior when choosing "delete" or "finish"

Create task with subtasks. Re-open it and complete subtask. Then click update. 

### Additional context

The behavior of "delete", "finish" and "update" is inconsistent with regards to what happens to the pop-up window.

Also, from a users perspective, it is unclear whether changes are saved only if the task has been "updated" manually. The function of this option is thus unclear. It overlaps with the close option. 

Either automatically save all changes (then, a specific update button seems unnecessary) and somehow include the option to discard changes if not wanted or include a "save" or "update" button to make sure changes are saved but alter function of the "close" option, potentially including the option to discard changes or keep them.

DasProffi:
closes #782


tasks/components/layout/TaskDetailView.tsx
```diff
@@ -366,9 +366,6 @@ export const TaskDetailView = ({
                   {translation.finish}
                 </Button>
               )}
-              <Button color="accent" onClick={() => updateTaskMutation.mutate(task)} disabled={!isValid}>
-                {translation.update}
```
MaxSchaefer, line 370: Also remove the now unused `update` translation key

MaxSchaefer (changes requested):

MaxSchaefer (approved):

PR #963, opened 2024-03-31
tasks as subtasks

### LINKED ISSUE #946: Tasks as Subtasks
### Describe the feature

Currently, a subtask on a task resolves around a `isDone` and `description` state. Therefore, a subtask is not a reference to another task. In the future, we will change that. A task can reference on another task in a parent-child-relation (e. g. Subtask).
We need to update the current task detail view to support this kind of subtask.

- The inline old-subtask view should represent a tree of new inline subtasks as planed in the wireframe
- "Add Subtask" should bring up our "New task" UI

Towards a final design, @DasProffi already did a wireframe. This should be enough to develop a first UI with our pre-existing components. Feel free to play around :)

fyi: This feature is currently not supported by the API, web makes the first move in this case.

https://github.com/helpwave/helpwave/issues/56

### Additional context

Wireframe
Bildschirmfoto 2024-03-27 um 12 20 05 AM

Current UI
Bildschirmfoto 2024-03-27 um 12 23 29 AM

DasProffi:
closes #946

### DIFF
--- a/lib/components/properties/PropertyBase.tsx
+++ b/lib/components/properties/PropertyBase.tsx
@@ -47,26 +47,21 @@ export const PropertyBase = ({
   const translation = useTranslation(defaultPropertyBaseTranslation, overwriteTranslation)
   const requiredAndNoValue = softRequired && !hasValue
   return (
-    <div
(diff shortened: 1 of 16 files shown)


DasProffi:
Out of date

PR #969, opened 2024-04-01
update Expandable and add FAQSection

### LINKED ISSUE #967: FAQ component
### Describe the feature

The FAQ component should accept a list of questions and answers. The answer can either be Markdown or another component. On-click on a question, the section should expand to see the answer. The caret-down icon changes to caret-up when the question is expanded. Also please ensure that the question is linkable via an anchor with h-Tags. (This could also be handles by the layout as well)

Image

On upcoming pages on our landing-page, we need a section to answer question. I decided on an FAQ structure for that. 

### Additional context

_See Figma_

DasProffi:
closes #967

Look at the storybook in layout

### DIFF
--- a/lib/components/Expandable.tsx
+++ b/lib/components/Expandable.tsx
@@ -7,9 +7,12 @@ export type ExpandableProps = PropsWithChildren<{
   label: ReactNode,
   icon?: (expanded: boolean) => ReactNode,
   initialExpansion?: boolean,
+  /**
+   * Whether the expansion should only happen when the header is clicked or on the entire component
+   */
+  clickOnlyOnHeader?: boolean,
   className?: string,
-  headerClassName?: string,
-  ref?: React.Ref<HTMLDivElement>
+  headerClassName?: string
 }>
 
 /**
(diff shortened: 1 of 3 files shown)


lib/components/Expandable.tsx
```diff
@@ -7,9 +7,12 @@ export type ExpandableProps = PropsWithChildren<{
   label: ReactNode,
   icon?: (expanded: boolean) => ReactNode,
   initialExpansion?: boolean,
+  /**
+   * Whether the expansion should only happen when the header is clicked or on the entire component
+   */
+  clickOnlyOnHeader?: boolean,
```
MaxSchaefer, line 13: Whats the reason behind introducing this new prop?

DasProffi: Normal `Expandable`s only expand when clicking the header as the content inside the expansion section might also use click events, and thus it would be confusing to have the expansion minimize on click. In other use cases like the FAQ Section on the other hand it would be confusing if the expansion only closes upon clicking the header.

MaxSchaefer (approved):

PR #1130, opened 2024-12-02
fix: adopt breaking changes of updates-svc

### LINKED ISSUE #1121: Adopt breaking changes of `updates-svc`
### Describe the chore

https://github.com/helpwave/services/pull/884 removes the `one-of` for `AggregateEvent` and `EntityEvent`. Its now just an `Event`.

MaxSchaefer:
closes #1121

### DIFF
--- a/tasks/package.json
+++ b/tasks/package.json
@@ -15,7 +15,6 @@
     "@headlessui/react": "1.7.19",
     "@helpwave/api-services": "workspace:*",
     "@helpwave/common": "workspace:*",
-    "@helpwave/proto-ts": "0.61.0-e1d7c0c.0",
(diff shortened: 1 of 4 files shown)


tasks/package.json
```diff
@@ -15,7 +15,7 @@
     "@headlessui/react": "1.7.19",
     "@helpwave/api-services": "workspace:*",
     "@helpwave/common": "workspace:*",
-    "@helpwave/proto-ts": "0.60.0-1191f09.0",
```
DasProffi, line 18: tasks should not depend on the protos. As it should be independent and only use them through the api-services. If this causes problems, we should open a separate issue.

DasProffi (approved):

PR #1132, opened 2024-12-03
Issue-1126: Slice TaskDetailView more SRP like

ChrisKulesza:
## Preparatory work for the processing of Issue-1126
- move `TaskDetailTranslation` and `translationData` to separate TS file
- move `TaskDetailViewSidebar` to separate TS file
- move `tasksDetails` to `TaskDetails`
- move `templateSidebar` to `TaskDetailViewTemplateSidebar`
- move `buttons` to `TaskDetailViewButtons`

## Motivation
- Structure code more legibly
- encapsulate technical differences between the individual components

none


tasks/components/layout/taskDetailView/TaskDetailViewSidebar.tsx
```diff
@@ -0,0 +1,182 @@
...
+import { Input } from '@helpwave/common/components/user-input/Input'
+import { formatDate } from '@helpwave/common/util/date'
+import { TimeDisplay } from '@helpwave/common/components/TimeDisplay'
+import type { TaskDetailViewTranslation } from './defaultTaskDetailViewTranslation'
+import { defaultTaskDetailViewTranslation } from './defaultTaskDetailViewTranslation'
+import { AssigneeSelect } from '@/components/selects/AssigneeSelect'
+import { TaskStatusSelect } from '@/components/selects/TaskStatusSelect'
+import { TaskVisibilitySelect } from '@/components/selects/TaskVisibilitySelect'
+
+type Props = {
```
DasProffi, line 24: If a component is exported, its Prop Type should be too, to allow for easy extensibility.
```suggestion
export type TaskDetailViewSidebarProps = {
```

tasks/components/layout/taskDetailView/TaskDetailViewSidebar.tsx
```diff
@@ -0,0 +1,182 @@
...
+  setTask: (task: TaskDTO) => void,
+  isCreating: boolean
+};
+
+export const TaskDetailViewSidebar = ({
+  overwriteTranslation,
+  task,
+  setTask,
+  isCreating
+}: PropsForTranslation<TaskDetailViewTranslation, Props>) => {
```
DasProffi, line 35:
```suggestion
}: PropsForTranslation<TaskDetailViewTranslation, TaskDetailViewSidebarProps >) => {
```

tasks/components/layout/taskDetailView/TaskDetailViewTemplateSidebar.tsx
```diff
@@ -0,0 +1,77 @@
...
+  wardId: string,
+  handleOnTileClick: (taskDTO: TaskDTO) => void
+};
+
+export const TaskDetailViewTemplateSidebar = ({
+  isCreating,
+  task,
+  wardId,
+  handleOnTileClick,
+}: Props) => {
```
DasProffi, line 26:
```suggestion
export type TaskDetailViewTemplateSidebarProps = {
  isCreating: boolean,
  task: TaskDTO,
  wardId: string,
  handleOnTileClick: (taskDTO: TaskDTO) => void
};

export const TaskDetailViewTemplateSidebar = ({
  isCreating,
  task,
  wardId,
  handleOnTileClick,
}: TaskDetailViewTemplateSidebarProps) => {
```

tasks/components/layout/taskDetailView/TaskDetailsButtons.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+export const TaskDetailsButtons = ({
+  overwriteTranslation,
+  task,
+  initialTaskId,
+  handleShowDeleteDialog,
+  handleFinishTask,
+  handleCreateTask
+}: PropsForTranslation<TaskDetailViewTranslation, Props>) => {
```
DasProffi, line 24:
```suggestion
export type TaskDetailsButtonProps = {
  task: TaskDTO,
  initialTaskId: string | undefined,
  handleShowDeleteDialog: (value: boolean) => void,
  handleFinishTask: (task: TaskDTO) => void,
  handleCreateTask: (task: TaskDTO) => void
};

export const TaskDetailsButtons = ({
  overwriteTranslation,
  task,
  initialTaskId,
  handleShowDeleteDialog,
  handleFinishTask,
  handleCreateTask
}: PropsForTranslation<TaskDetailViewTranslation, TaskDetailsButtonProps>) => {
```

DasProffi (commented):
The changes look good and improve the readability a lot (especially moving the subcomponents outside the `TaskDetailView` Component function into separate components). I am unsure whether using multiple files is the best solution in regard to finding the right component to change (due to similar naming, it is unclear which is the main file).

The previously used **single file component design**, makes sure only the main component is exported, while now every subcomponent is exported too and thus within the search space of the app, although its use is only inside the `TaskDetailView`.

Either 1) declare each subcomponent its own reusable component, or 2) follow the **single file component design**.

Changes for 1):
- make the props of the component public with an export and name them after the component
- use and declare individual translation within each component (in this case, I would see it as optional)

Changes for 2):
Move the subcomponents and Translations back inside the `TaskDetailView.tsx` and only export the `TaskDetailView` component.

Apart from that, fix the linting for the CI.

DasProffi:
> Personally, I am flexible regarding how the Helpwave application is structured. In my day-to-day work, I am accustomed to organizing by features as well as following the SRP approach. If the preference here at Helpwave is to handle this with a large component approach, I am happy to implement it that way.
> 
> I have adapted this because working with a 450+ line component was a little difficult.
> 
> Could you please provide a clear direction? Thank you.

Keep the subcomponents  `TaskDetailViewSidebar`, `TaskDetails`, `TaskDetailViewTemplateSidebar`, `TaskDetailViewButtons`, but move them into a single file (approach 2 from my previous message). Only export the `TaskDetailsView` and have the `TaskDetailTranslation` also within that file.

DasProffi:
Will be done in #1143

PR #1147, opened 2024-12-26
add getBeds query to api services

### LINKED ISSUE #1103: Remove usage of `GetBedsByRooms` and use `GetBeds` instead
### Describe the chore

We deprecated the `GetBedsByRooms` endpoint in services and added an optional `room_id` to the `GetBeds` endpoint

DasProffi:
closes #1103

### DIFF
--- a/api-services/mutations/tasks/bed_mutations.ts
+++ b/api-services/mutations/tasks/bed_mutations.ts
@@ -2,7 +2,7 @@ import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import {
   CreateBedRequest,
   DeleteBedRequest,
-  GetBedRequest,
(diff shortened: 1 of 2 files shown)


MaxSchaefer:
@DasProffi `GetBedsByRoom` was already removed? Checkout the offline service. Search `GetBedsByRoom` across the codebase.

DasProffi:
> @DasProffi `GetBedsByRoom` was already removed? Checkout the offline service. Search `GetBedsByRoom` across the codebase.

Yes, therefore the new `GetBedsQuery`  was implemented where `roomId` is an optional parameter to only query beds of a certain room.

The method still is overwritten in the `BedOfflineServicePromiseClient` in order to not forward it to an actual backend. I have marked it as deprecated, and we can remove it once the protos of services remove it.

PR #1150, opened 2024-12-27
update linting

### LINKED ISSUE #1100: Replace empty `.then()` with `.catch()` and error logging
### Describe the chore

We have many empty `.then()`'s all around the place. We did this to satisfy our linter but we should use `.catch()` with error logging instead because an empty `.then()` omits every error that might be thrown.

- [ ] Disallow empty `.then()` via the linter
- [ ] Replace empty `.then()` with `.catch()` and error logging

DasProffi:
closes #1100

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,20 @@ This repository is split up into multiple subprojects using [pnpm](https://pnpm.
 - library of helpwave (see [lib](/lib))
 
 ## Getting Started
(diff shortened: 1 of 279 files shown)


MaxSchaefer:
I am poking around locally, when running `pnpm run lint` inside `tasks/` I get a tsc error. Can you reproduce this?

MaxSchaefer (approved):

